PR #2, opened 2016-07-16
examples for nvcxx

vchuravy:
All this is in prototypical stage and need better integration.

`saxpy.jl` is functional and `curand.jl` still needs a bit of work.

### DIFF
--- a/examples/nvcxx/curand.jl
+++ b/examples/nvcxx/curand.jl
@@ -1,0 +1,82 @@
+###
+# Using Cxx.jl to compile a C++ cuda kernel and then call it from Julia. 
+###
+using Cxx, CUDAdrv, CUDAnative
+
+# Figure out if the CPU is set to the correct value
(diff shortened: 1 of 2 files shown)


vchuravy:
The examples require LLVM 3.8 and the current master of Cxx.jl

vchuravy:
I extended the examples to contain the calls to CUDAdrv and CUDAnative

vchuravy:
Updated the curand demo. Needs #3.

maleadt:
I wouldn't close this, it contains the necessary boilerplate for setting-up Cxx and it's definitely where we want to end up, it's just that for now it's impossible to make this perform well (lacking some "pointer to otherwise non-escaping stack object" primitive)...

vchuravy:
I closed this because all of this machinery now lives in https://github.com/JuliaGPU/CUDAnativelib.jl, but we can keep this since it is a more minimal example.

PR #36, opened 2017-02-01
Replace cudaconvert with simple pure funtion.

maleadt:
Generated functions are fragile, and we were doing stuff we shouldn't (calling overloaded functions and calling its methods after the generated function was defined). This broke the package if precompilation was disabled, but even with precompilation enabled the generated function behind `@cuda` threw errors because during first inference-time execution `cudaconvert(DevicePtr{T})` wasn't defined. This also slowed down the kernel launch sequence.

Ref JuliaLang/julia#19942

### DIFF
--- a/src/CUDAnative.jl
+++ b/src/CUDAnative.jl
@@ -15,12 +15,11 @@ isfile(ext) || error("Unable to load $ext\n\nPlease re-run Pkg.build(\"CUDAnativ
 include(ext)
 
 include("util.jl")
-include("types.jl")
 
 include("jit.jl")
-include("execution.jl")
-include("device/intrinsics.jl")
 include("device/array.jl")
+include("device/intrinsics.jl") # these files contain generated functions,
+include("execution.jl")         # so should get loaded last (JuliaLang/julia#19942)
 
 
 function __init__()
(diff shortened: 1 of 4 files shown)


vchuravy:
This means we won't be able to define convert functions outside CUDAnative?

maleadt:
Correct... I think the situation is actually much worse. Quoting @vtjnash:

> it’s wrong to mark a function `Base.@pure` if it can be overloaded
> but I also see a lot of calls to `size`, `promote_op`, etc. from `@generated` functions
> but `@generated` functions can only call `pure` functions

... but the generated function backing `@cuda` is doing [much much more](https://github.com/JuliaGPU/CUDAnative.jl/blob/0658dc09c0362d88f7997e2a5149fe99ba19a738/src/execution.jl#L195-L232).

The reason I implemented all this in a generated function is to emit only the code necessary to process (convert, upload, download) the kernel arguments based on their types, resulting in a very efficient kernel launch sequence.

Maybe it would be possible (ie. as fast) to implement this with regular functions, using a combination of `@inline` and relying on the compiler to propagate type information and specialize code? I'm afraid it will not, because we're doing pretty complex stuff, eg.:

```julia
@cuda (1,1) kernel_reduce(+, some_array)

# the generated function emits a specialized cudacall like:
cudacall(kernel_reduce, (CuDeviceArray{T},), some_array)
# ditching the + ghost argument

# if we were to do this at runtime, we'd probably need to
kernel_args = Any[]
kernel_argtyps = Type[]
for arg in args
  if !isghost(arg)
    # call cudaconvert (eg. CuArray -> CuDeviceArray)
    # allocate GPU objects
    # push to arrays from above
  end
end
cudacall(kernel_reduce, tuple(kernel_argtyps...), kernel_args...)
# which will probably be slow?
```

PR #47, opened 2017-02-25
Extend shared memory to arbitrary (isbits) types

SimonDanisch:
@vchuravy and I figured out this way to deal with arbitrary element types.
It passes the test and seems to yield the correct results in my kmeans demo!
But I get world age errors in kmeans, disabling type inference. Not sure where they originate from, since the tests don't trigger it! Maybe it's something I do in GPUArrays!


src/device/intrinsics.jl
```diff
@@ -331,10 +331,26 @@ macro cuStaticSharedMem(typ, dims)
...
 function emit_static_shmem{N}(id::Integer, jltyp::Type, shape::NTuple{N,Int})
-    if !haskey(llvmtypes, jltyp)
-        error("cuStaticSharedMem: unsupported type '$jltyp'")
+    var = Symbol(:@shmem, id)
+    len = prod(shape) * sizeof(jltyp)
+    align = Base.Threads.alignment(jltyp)
+    return quote
+        Base.@_inline_meta
+        ptr = Base.llvmcall(
+            ($"""$var = internal addrspace(3) global [$len x i8] zeroinitializer, align $align""", # Question alignment correct
```
vchuravy: That comment should be removed

src/device/intrinsics.jl
```diff
@@ -331,10 +331,26 @@ macro cuStaticSharedMem(typ, dims)
...
 end
 
+
+
 function emit_static_shmem{N}(id::Integer, jltyp::Type, shape::NTuple{N,Int})
-    if !haskey(llvmtypes, jltyp)
-        error("cuStaticSharedMem: unsupported type '$jltyp'")
+    var = Symbol(:@shmem, id)
+    len = prod(shape) * sizeof(jltyp)
+    align = Base.Threads.alignment(jltyp)
```
vchuravy: We are asking the GC here what alignment it would use for the type. I don't recall if this is only correct for isbits or if it holds in general but it gives reasonable defaults.

src/device/intrinsics.jl
```diff
@@ -493,7 +509,22 @@ for typ in (Int64, UInt64, Float64)
     end
 end
 
-
+@generated function shfl_down{T}(val::T, srclane::Integer, width::Integer = Int32(32))
```
vchuravy: This should probably not be part of this PR.

SimonDanisch: oh, haha! True story...

src/device/intrinsics.jl
SimonDanisch: I guess, ptx codegen won't support anything but isbits anyways?

maleadt:
Looks great, thanks. Maybe add some tests actually writing to/reading from memory such that we can detect invalid accesses if we were to run the test-suite under `cuda-memcheck`.

Should we also use that `Base.Threads.align` for other alignment, like the other `emit_static_shmem` case, or the [ptrset and ptrref instinrics](https://github.com/JuliaGPU/CUDAnative.jl/blob/ec0bd75bf71c67a455318fc8b84784e10ed5b41b/src/device/array.jl#L42)? cc @vchuravy

vchuravy:
IIRC `Base.Threads.alignment` gives the alignment that the GC guarantees for a struct (cc: @yuyichao). My reasoning was that should be the same as the necessary alignment for a struct, so we should use it in other places as well.

yuyichao:
> My reasoning was that should be the same as the necessary alignment for a struct.

Not really.

maleadt:
Extended to dynamic memory, and cleaned up a bit. Still lacks some correctness tests, but should be good to go for now.

Issue #25, opened 2017-01-11
Int64 literals vs Int32 constants: avoid conversions & checks


maleadt:
Many constants in CUDA world are 32-bit, eg. the warp-size, thread or block IDs and dimensions, etc. We don't promote these to Int64 in order to avoid conversions when doing math on them, however it might be equally expensive not to do so because of conversions when doing math with literals.

For example, take the following idiomatic code:

```julia
function reduce_warp{F<:Function,T}(op::F, val::T)::T
    offset = CUDAnative.warpsize() ÷ 2
    while offset > 0
        val = op(val, shfl_down(val, offset))
        offset ÷= 2
    end
    return val
end
```

`warpsize` yields an Int32, but gets converted and promoted to Int64 because of the `÷ 2`. This in turn causes `shf_down` which takes an Int32 do convert it back, including an exactness check + exception (trap):

```llvm
julia> CUDAnative.code_llvm(reduce_warp, (typeof(+), Int32))

define i32 @julia_reduce_warp_62748(i32) local_unnamed_addr # {
top:
  %1 = tail call i32 @llvm.nvvm.read.ptx.sreg.warpsize()
  %2 = icmp slt i32 %1, 2
  br i1 %2, label %L23, label %if.preheader

if.preheader:                                     ; preds = %top
  %3 = lshr i32 %1, 1
  %4 = zext i32 %3 to i64
  br label %if

if:                                               ; preds = %if.preheader, %pass2
  %val.03 = phi i32 [ %9, %pass2 ], [ %0, %if.preheader ]
  %offset.02 = phi i64 [ %10, %pass2 ], [ %4, %if.preheader ]
  %sext = shl i64 %offset.02, 32
  %5 = ashr exact i64 %sext, 32
  %6 = icmp eq i64 %5, %offset.02
  br i1 %6, label %pass2, label %fail1

L23.loopexit:                                     ; preds = %pass2
  br label %L23

L23:                                              ; preds = %L23.loopexit, %top
  %val.0.lcssa = phi i32 [ %0, %top ], [ %9, %L23.loopexit ]
  ret i32 %val.0.lcssa

fail1:                                            ; preds = %if
  tail call void @llvm.trap()
  unreachable

pass2:                                            ; preds = %if
  %7 = trunc i64 %offset.02 to i32
  %8 = tail call i32 @llvm.nvvm.shfl.down.i32(i32 %val.03, i32 %7, i32 31)
[...]

PR #50, opened 2017-03-22
WIP: Introduce at-narrow to narrow literals in expressions.

maleadt:
May need an NSFW tag.

Ref #25

### DIFF
--- a/src/CUDAnative.jl
+++ b/src/CUDAnative.jl
@@ -11,6 +11,7 @@ isfile(ext) || error("Unable to load $ext\n\nPlease run Pkg.build(\"CUDAnative\"
 include(ext)
 
 include("jit.jl")
+include("device/util.jl")
 include("device/array.jl")
 include("device/intrinsics.jl") # these files contain generated functions,
 include("execution.jl")         # so should get loaded quite late (JuliaLang/julia#19942)
--- a/src/device/util.jl
+++ b/src/device/util.jl
@@ -15,15 +15,15 @@ macro narrow(width, ex)
 end
 
 macro narrow32(ex)
-    quote
+    esc(quote
         @narrow 32 $ex
-    end
+    end)
 end
 
 narrow_code(width, ex) = ex
 function narrow_code(width, ex::Expr)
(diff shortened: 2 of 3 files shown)


maleadt:
Hmm.
```julia
julia> dump(@narrow 1000÷-2)
UInt16 65036

julia> dump(1000÷-2)
Int64 -500
```

maleadt:
wat
```julia
julia> dump(UInt16(1000)÷-2)
UInt64 18446744073709551116

julia> dump(UInt16(1000)÷-UInt64(2))
UInt64 0

julia> dump(UInt16(1000)÷-UInt32(2))
UInt32 0

julia> dump(UInt16(1000)÷-UInt16(2))
UInt16 0

julia> dump(UInt16(1000)÷-UInt8(2))
UInt16 3
```

OK maybe I shouldn't be doing this.

maleadt:
Hmm, unsigned behaviour doesn't seem all that reliable, let alone at different bit widths:
````
julia> dump(UInt8(255)+1)
Int64 256

julia> dump(UInt16(255)+1)
Int64 256

julia> dump(UInt32(255)+1)
Int64 256

julia> dump(UInt64(255)+1)
UInt64 256
```

Maybe I should just specialize `narrow` to make literals 32-bit and converts to Int/UInt without width use 32-bit variants.

SimonDanisch:
I wonder how helpful this is! The main reason why I personally would like to have better behaviour is, to make it easier to work with user defined function that don't even know of this problem. If they don't know of the problem, they obviously won't use a narrow macro... And people who know of this, can easily sprinkle UInt32(x) around, no?
Or am I misinterpreting the purpose of this PR?

maleadt:
> And people who know of this, can easily sprinkle UInt32(x) around, no?

You're right, I mostly intended it to be a quick hack to keep code readable (while porting Rodinia benchmarks). If it turns out to be reliable and doesn't break too much code, we could however consider using it as part of `@cuda`. But #25 definitely needs some more thought than just this band-aid.

maleadt:
https://github.com/stevengj/ChangePrecision.jl

PR #52, opened 2017-04-03
WIP/RFHelp: kernel precompilation

maleadt:
In an attempt to speed up CUDAnative, I've implemented a precompile-like compile cache (opt-in). Turns out that doesn't really speed much up... Not sure how to proceed though.

What is the current way of speeding up package load times? I figured "SnoopCompile + precompile.jl", but that lists some really strange entries:

```

julia> SnoopCompile.@snoop "/tmp/vadd.csv" begin
       include("examples/vadd.jl")
       end
Launching new julia process to run commands...
done.

julia> data = SnoopCompile.read("/tmp/vadd.csv")

 13737776  …  "Base.Type(Type{Ref{Ptr{LLVM.API.LLVMOpaqueValue}}}, Array{Any, 1})"                                                                                                                                                                                                                                                                
 24132971     "Base.unsafe_convert(typeof(Base.unsafe_convert), Type{Ptr{Ptr{LLVM.API.LLVMOpaqueType}}}, Base.RefArray{Ptr{LLVM.API.LLVMOpaqueType}, Array{Ptr{LLVM.API.LLVMOpaqueType}, 1}, Any})"                                                                                                                                               
 37450972     "Base.Type(Type{Ref{Ptr{LLVM.API.LLVMOpaqueType}}}, Array{Any, 1})"                                                                                                                                                                                                                                                                 
 91889685  …  "Base.ht_keyindex(typeof(Base.ht_keyindex), Base.Dict{UInt32, Type{T} where T}, UInt32)" 
```

And as far as the documentation goes, we can't precompile functions in other modules (here, Base)?

### DIFF
--- a/examples/vadd.jl
+++ b/examples/vadd.jl
@@ -11,6 +11,9 @@ end
 dev = CuDevice(0)
 ctx = CuContext(dev)
 
+CUDAnative.precompile(kernel_vadd,
+                      (CuDeviceArray{Float32,2},CuDeviceArray{Float32,2},CuDeviceArray{Float32,2}))
+
 dims = (3,4)
 a = round.(rand(Float32, dims) * 100)
 b = round.(rand(Float32, dims) * 100)
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -217,6 +217,99 @@ const func_cache = Dict{UInt, CuFunction}()
     end
 end
 
+# alternatively, users can explicitly precompile kernels
+if is_linux()
+    const func_cache_dir = joinpath(get(ENV, "XDG_CACHE_HOME", joinpath(homedir(), ".cache")),
+                                    "CUDAnative.jl")
+else
+    const func_cache_dir = joinpath(tempdir(), "CUDAnative.jl")
+end
+function precompile(func::ANY, types::ANY)
+    @assert typeof(types) <: Tuple
+    precomp_key = hash(tuple(typeof(func), types...))
+
+    # details about the code
+    meth = which(func, types)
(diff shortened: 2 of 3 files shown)


maleadt:
> And as far as the documentation goes, we can't precompile functions in other modules (here, Base)?

Confirmed that one, after adding `precompile(Base.ht_keyindex, (Base.Dict{UInt32, Type{T} where T}, UInt32))` to `CUDAnative.jl` and recompiling the cache file:

```julia
julia> ccall(:jl_dump_compiles, Void, (Ptr{Void},), STDOUT)
julia> using CUDAnative
julia> precompile(Base.ht_keyindex, (Base.Dict{UInt32, Type{T} where T}, UInt32))
539607  "Base.ht_keyindex(typeof(Base.ht_keyindex), Base.Dict{UInt32, Type{T} where T}, UInt32)"
```

MikeInnes:
You may not need to go outside of the usual precompilation mechanism for this. If you cache compiled kernels in a per-module `ObjectIdDict`, precompilation should just store that in the usual ji files. This is essentially what the doc system does as well.

maleadt:
Going to close this, as most of the overhead is due to Julia compiling the packages, instead of us compiling kernels. For example, on the reasonably complex kernels from `reduce.jl`, cold compilation takes 12.8s, but hot compilation only 0.14s (cached compilation 0.034ms).

MikeInnes:
I'm sure it won't take too long before people are compiling tens of cuda kernels in a more complex program. Nevertheless, it makes sense to deal with this as it comes up of course.

PR #54, opened 2017-04-20
DOA: Small event-based profiler.

maleadt:
Example usage;

```julia
using CUDAdrv, CUDAnative

function kernel_vadd(a, b, c)
    i = (blockIdx().x-1) * blockDim().x + threadIdx().x
    c[i] = a[i] + b[i]

    return nothing
end

const dev = CuDevice(0)
const ctx = CuContext(dev)

function main()
    dims = (3,4)
    a = round.(rand(Float32, dims) * 100)
    b = round.(rand(Float32, dims) * 100)

    d_a = CuArray(a)
    d_b = CuArray(b)
    d_c = similar(d_a)

    len = prod(dims)
    @cuda (1,len) kernel_vadd(d_a, d_b, d_c)
    Array(d_c)
end

# CUDAnative.Profile.init(nvprof=true)

CUDAnative.@profile begin
    main()
end

CUDAnative.Profile.print()

destroy(ctx)
```

cc @cfoket

### DIFF
--- a/src/CUDAnative.jl
+++ b/src/CUDAnative.jl
@@ -11,6 +11,7 @@ isfile(ext) || error("Unable to load $ext\n\nPlease run Pkg.build(\"CUDAnative\"
 include(ext)
 
 include("jit.jl")
+include("profile.jl")
 include("device/array.jl")
 include("device/intrinsics.jl") # these files contain generated functions,
 include("execution.jl")         # so should get loaded quite late (JuliaLang/julia#19942)
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -129,7 +129,7 @@ function emit_allocations(args, codegen_types, call_types)
     return kernel_allocations, args
 end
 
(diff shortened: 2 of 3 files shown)


maleadt:
Not sure whether this should be called a "profiler". Real profiling would require CUPTI.

maleadt:
This approach is probably DOA. There's an overhead that cannot be avoided by calling into the CUDA API and copying over arguments to the GPU. This also introduces quite some variability. For example, with Rodinia's hotspot the main kernel executes, according to this profiler, in 156.4 +/- 3 us, while `nvprof` reports 111.75 +/- 0.13 us. Ignoring the constant overhead, which also happens with CUDA C code (albeit typically a little bit less Int32 vs Int64 args etc), but the variability is a bummer.

I'm probably going to leave this here for inspiration, in case somebody wants to revisit this based on CUPTI.

PR #55, opened 2017-04-21
use CUDA_HOME to search for libdevice

vchuravy:


maleadt:
Too bad we need access to the toolkit (after removing the CUDA C example for exactly this reason).

Maybe we should redistribute? Libdevice files are [listed as redistributable](http://docs.nvidia.com/cuda/eula/index.html#attachment-a) in the CUDA EULA, but [requires](http://docs.nvidia.com/cuda/eula/#redistribution-rights) `each Customer's written or clickwrap agreement to the license terms under a written, legally enforceable agreement that has the effect of protecting the SOFTWARE and the rights of NVIDIA under terms no less restrictive than this Agreement`...

vchuravy:
In theory we could query once during a Pkg.build for the user to accept the
agreement (or accept it beforehand through an environment flag).
But doesn't the `libdevice` library change with different versions of cuda?

It could also be the responsibility of CUDAdrv to provide these paths,
instead of having to reimplement the logic here.

maleadt:
`libdevice` changes, but shouldn't be incompatible.

I'd suggest honouring not only CUDA_HOME, but also CUDA_ROOT and CUDA_PATH (cfr. [CUDArt.jl](https://github.com/JuliaGPU/CUDArt.jl/blob/3f49f470fb5f9e3b1ca96aad0f195ce342316d6f/deps/build.jl#L50-L62)), as well as embedding the resulting `libdevice_dir` in `ext.jl`. The package shouldn't be loadable if we're missing `libdevice`.

I'll have a look next week, as I need to clean-up CUDAnative's `build.jl` anyway (including @musm's fixes to CUDArt/CUDAdrv's build scripts).

PR #56, opened 2017-04-21
Add tiny profiler to enable the CUDA profiler upon kernel launch.

maleadt:


maleadt:
Relies on a new tag for CUDAdrv.

PR #57, opened 2017-04-24
Uniformize build script with other CUDA packages.

maleadt:


maleadt:
This is also preparatory work towards a build-time alternative to #55. Will be adding here.

PR #60, opened 2017-05-11
Enable generation of FMA instructions.

cfoket:
This patch passes `--nvptx-fma-level=1` to LLVM's NVPTX backend to mimic the default behavior of `nvcc` with respect to the generation of FMA instructions. Note that `nvcc`'s `--fmad` option defaults to `true`.


maleadt:
Do we need `nvptx-mad-enable` too?

maleadt:
Oops, forgot about this.

PR #61, opened 2017-05-19
WIP: Automatic replacements of Base intrinsics with CUDAnative alternatives

SimonDanisch:
This is the script we discussed in https://github.com/JuliaGPU/GPUArrays.jl/issues/17!
It needs Sugar 0.2.0, which is currently not tagged.

Usage:
```Julia
function deep_recurse(x)
    sin(sqrt(x))
end
function recursive_test(t)
    cos(t) + deep_recurse(t)
end
function haversine{T <: Float32}(lat1::T, lon1::T, lat2::T, lon2::T, radius::T)
    c1 = cospi(lat1 / 180.0f0)
    c2 = cospi(lat2 / 180.0f0)
    dlat = lat2 - recursive_test(lat1)
    dlon = lon2 - lon1
    d1 = sinpi(dlat / 360.0f0)
    d2 = sinpi(dlon / 360.0f0)
    t = d2 * d2 * c1 * c2
    a = d1 * d1 + t
    c = 2.0f0 * asin(min(1.0f0, sqrt(a)))
    return radius * c
end

cu_haversine, rewritten = rewrite_for_cudanative(haversine, Tuple{Float32, Float32, Float32, Float32, Float32})
```

Would be nice if you could do the real integration, since I don't have access to a CUDA device right now...

### DIFF
--- a/src/rewrite_intrinsics.jl
+++ b/src/rewrite_intrinsics.jl
@@ -1,3 +1,6 @@
+import Sugar
+using Sugar: LazyMethod
+
 function rewrite_expr_cu(m::LazyMethod, expr)
     was_rewritten = false
     body = first(Sugar.replace_expr(expr) do expr
@@ -6,11 +9,7 @@ function rewrite_expr_cu(m::LazyMethod, expr)
             if head == :call
                 func = args[1]
                 types_array = map(args[2:end]) do x
-                    t = Sugar.expr_type(m, x)
-                    if t == Any
-                        error("Found type any in Expr: $expr. Make sure your code is type stable")
-                    end
-                    t
+                    Sugar.expr_type(m, x)
                 end
                 types = (types_array...,)
                 f = Sugar.resolve_func(m, func)
@@ -50,8 +49,15 @@ function rewrite_for_cudanative(f, types)
     m = LazyMethod((f, types))
     # if is a Julia intrinsic, stop
     Sugar.isintrinsic(m) && return f, false
+    isa(Sugar.getfunction(m), DataType) && return f, false
     # otherwise go through the source and rewrite function calls recursevely the source!
-    expr = Expr(:block, Sugar.get_ast(code_typed, m.signature...)...)
+   try
+        expr = Sugar.sugared(code_typed, m.signature...)
+    catch e
+        # TODO warn or explicitely filter out errors that are expected?
+        # E.g. it can't get the ast for some stuff like Base.cconvert(DataType, x)
+        return f, false
+    end
     body, was_rewritten = rewrite_expr_cu(m, expr)
     if was_rewritten
         rewritten_func_expr = Sugar.get_func_expr(m, body, gensym(string("cu_", fsym)))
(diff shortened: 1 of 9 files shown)


SimonDanisch:
https://github.com/JuliaLang/METADATA.jl/pull/9400

SimonDanisch:
Got around to integrate this...
I think it might be a bit brittle. It passes tests (locally) but I'm sure people will find functions, that Sugar can't handle correctly. It's maybe fine as a stop gap solution... If you don't like this, I'm fine with integrating it into GPUArrays for now, since the same functions should fail for the OpenCL backend, so it would be at least consistent ;)

SimonDanisch:
oh i see, 0.7 changed stuff already :D

SimonDanisch:
Lol it passes tests, but it seems like it doesn't actually replace the functions correctly

src/rewrite_intrinsics.jl
```diff
@@ -0,0 +1,68 @@
...
+end
+
+"""
+Replaces recursively Julia Base functions which are defined as intrinsics in CUDAnative
+Returns the resulting function and a bool indicating, wether the function was rewritten.
+"""
+function rewrite_for_cudanative(f, types)
+    fsym = Symbol(f)
+    # if defined in CUDAnative, but doesn't match the function passed in here
+    # The probability is very high, that this is a cudanative intrinsic
```
vchuravy: How about we move all intrinsics into a separate module so that this we can actually restrict that check to that?

maleadt: A `CUDA` submodule would make code look nice, eg. `CUDA.sin`. Kinda hard to move with a deprecation in place though, unless we add a hack to `@cuda` scanning for them.

SimonDanisch:
I run into quite a lot of inference issues with this... Still need to figure out what really goes wrong here. 
I need to put in on hold for now, but I'll try to come up with a minimal example to investigate things better...

maleadt:
I just came across [AppleAccelerate.jl](https://github.com/JuliaMath/AppleAccelerate.jl) which has:

```julia
AppleAccelerate.@replaceBase sin cos tan
@time sin(X) # will use AppleAccelerate methods for vectorised operations
```

SimonDanisch:
That seems pretty unrelated? I mean, if we could have a version that works both on CPU and CUDAnative, than we could also just overload e.g. `sin`...But we don't, no?

maleadt:
Well, if you want to write generic source-code then yes this is unrelated. But seeing how it's apparently pretty hard to get right, and isn't even semantically correct, a solution like `@replaceBase` (possibly scoped) might be a viable alternative to cut down on the necessary namespacing.

Also, I was just pointing out other people are dealing with similar issues too 🙂

vchuravy:
What is our current status/thinking on this? Are we waiting for Cassette.jl?

maleadt:
Yeah. Despite @SimonDanisch's excellent work on Sugar I don't feel comfortable enabling it (it can't handle most code yet, eg. the test suite), so we'll probably just wait on Cassette.

maleadt:
I don't think this is going to happen with Sugar.jl (which doesn't seem very maintained...), so I'm going to close this. In the meantime, we have much better error reporting when using Base intrinsics, as well as broadcast rewriting, so the situation has improved already.

PR #65, opened 2017-06-01
Add Vector/Matrix type aliases for CuDeviceArray

maleadt:


maleadt:
Ah right, requires a new tag on CUDAdrv of course.

examples/reduce/reduce.jl
```diff
@@ -12,7 +12,7 @@ using CUDAdrv, CUDAnative
 #
 
 # Reduce a value across a warp
-@inline function reduce_warp{F<:Function,T}(op::F, val::T)::T
+@inline function reduce_warp(op::Function, val::T)::T where {T}
```
vchuravy: Doesn't `{F<:Function}` force a specialisation for the function?

maleadt: I don't think it matters: https://docs.julialang.org/en/latest/manual/style-guide.html#Don%27t-use-unnecessary-static-parameters-1

PR #67, opened 2017-06-02
Automatic recompilation of dependent kernels.

maleadt:
```
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.6.0-rc2.0 (2017-05-18 02:31 UTC)
 _/ |\__'_|_|_|\__'_|  |  
|__/                   |  x86_64-unknown-linux-gnu

julia> using CUDAdrv, CUDAnative

julia> dev = CuDevice(0)
CUDAdrv.CuDevice(0, 0)

julia> ctx = CuContext(dev)
CUDAdrv.CuContext(Ptr{Void} @0x0000000000fa9520, true, true)

julia> foo() = @cuprintf("old")

julia> @cuda (1,1) foo(); synchronize()
old

julia> foo() = @cuprintf("new")

julia> @cuda (1,1) foo(); synchronize()
new
```

Needs test. Better way would be to have an inferable `const age = current_method_age` and introduce a function barrier, but this is better than nothing.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -139,6 +139,8 @@ function emit_cudacall(func, dims, shmem, stream, types, args)
     end
 end
 
+world_age() = ccall(:jl_get_tls_world_age, UInt, ())
+
 
 #
 # @cuda macro
@@ -177,7 +179,7 @@ macro cuda(config::Expr, callexpr::Expr)
 end
 
 # Compile and execute a CUDA kernel from a Julia function
-const func_cache = Dict{UInt, CuFunction}()
+const compilecache = Dict{UInt, CuFunction}()
 @generated function generated_cuda{F<:Core.Function,N}(dims::Tuple{CuDim, CuDim}, shmem, stream,
                                                        func::F, args::Vararg{Any,N})
     arg_exprs = [:( args[$i] ) for i in 1:N]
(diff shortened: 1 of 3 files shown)


maleadt:
Original, alternative approach 1:

```julia
function method_age(f, tt)
    for m in Base._methods(f, tt, 1, typemax(UInt))
        return m[3].min_world
    end
    return -1
end
```

Pretty slow though, so on advice of Jameson I boiled it down to a smaller set of calls to the compiler:

```
function method_lookup_by_type(f::ANY, t::ANY, world=typemax(UInt))
    t = Base.to_tuple_type(t)
    t = Tuple{isa(f,Type) ? Type{f} : typeof(f), t.parameters...}
    return ccall(:jl_method_lookup_by_type, Ref{Core.MethodInstance}, (Any, Any, Cint, Cint, Cint, Csize_t), typeof(f).name.mt, t, 0, 0, 1, world)
end

function method_age(f, tt)
    return method_lookup_by_type(f, tt).min_world
end
```

Still too slow though.

MikeInnes:
@jrevels and @vtjnash discussed this in a different context and were thinking about adding compiler support, I think.

PR #68, opened 2017-06-04
Add documentation infrastructure

maleadt:


maleadt:
CI failure is unrelated, due to incompatibility with current master.

PR #77, opened 2017-07-12
Use a default, global context for the first device.

maleadt:
Simplifies code quite a bit. Mimics what CUDA does.

Maybe we should tie into the primary context business though, but then everyone's GC/refcount logic has to be aware of one another for it to work properly. Didn't really feel like doing that today. People should just use explicit contexts if they want to have control over resources, I guess.


Evizero:
very convenient thanks

PR #80, opened 2017-07-14
Fine-grained recompilation of dependent kernels.

maleadt:
Fixes #67.


maleadt:
Slightly lowers launch performance due to second `Dict` look-up (~1us), but that's not too bad.

PR #83, opened 2017-07-24
Align memory accesses according to the datatype.

maleadt:
After #82, shared memory is aligned according to the datatype alignment rules. At the same time, the documentation for `cuMemAlloc` states `[the] allocated memory is suitably aligned for any kind of variable`. But some online sources state a 256-byte alignment from old documentation (`Any address of a variable residing in global memory or returned by one of the memory allocation routines from the driver or runtime API is always aligned to at least 256 bytes`), others seem to refer `textureAlignment`, while the PTX documentation for global loads states `[global] memory instructions support reading or writing words of size equal to 1, 2, 4, 8, or 16 bytes`...

So I'm not sure whether this is correct. Running our testsuite under `cuda-memcheck` now.

### DIFF
--- a/src/device/util.jl
+++ b/src/device/util.jl
@@ -132,3 +132,7 @@ function escape_llvm_string(io, s::AbstractString, esc::AbstractString)
     end
 end
 escape_llvm_string(s::AbstractString) = sprint(endof(s), escape_llvm_string, s, "\"")
+
+
+Base.@pure alignment(::Type{T}) where {T} =
+    convert(Int, ccall(:jl_alignment, Cint, (Csize_t,), sizeof(T)))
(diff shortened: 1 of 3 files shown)


maleadt:
`cuda-memcheck` didn't report any misalignments.

@vchuravy, care to look over this? You've been spending time with GC alignment issues, does this look OK or am I confusing different aspects here?

src/device/util.jl
```diff
@@ -132,3 +132,7 @@ function escape_llvm_string(io, s::AbstractString, esc::AbstractString)
     end
 end
 escape_llvm_string(s::AbstractString) = sprint(endof(s), escape_llvm_string, s, "\"")
+
+
+Base.@pure alignment(::Type{T}) where {T} =
+    convert(Int, ccall(:jl_alignment, Cint, (Csize_t,), sizeof(T)))
```
vchuravy: You might want to use `jl_datatype_alignment` here or ask LLVM for its preferred alignment. `jl_alignment` is more about GC memory alignment guarantees.

PR #85, opened 2017-07-30
More general conversions

MikeInnes:
This patch lifts the `CuArray` -> `CuDeviceArray` and similar conversions out of the generated function, which makes it possible for downstream libraries to hook into the system (e.g. for other array types).

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -9,63 +9,17 @@ using Base.Iterators: filter
 # Auxiliary
 #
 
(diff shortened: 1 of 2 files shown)


maleadt:
Looks great, thanks. Some nice simplifications in there as well -- always healthy to have a second pair of eyes going through code :-) I'll be folding it into #84, because of some conflicts with the `DevicePtr` refactoring.

MikeInnes:
Works for me, thanks :+1:

PR #88, opened 2017-08-03
Detect non-specsig functions early

maleadt:


maleadt:
@one-more-minute: should detect your non-specsig case from yesterday:

```
julia> using CUDAnative
INFO: Recompiling stale cache file /home/tbesard/Work/Julia-CUDA/.julia/0.6-release/lib/v0.6/CUDAnative.ji for module CUDAnative.

julia> foo(x...) = nothing
@foo (generic function with 1 method)

julia> @cuda (1,1) foo(1)
ERROR: ArgumentError: invalid device function foo(Int64): is a varargs function
```

MikeInnes:
Great – is there any chance the vararg could be lifted into a tuple argument in future?

maleadt:
Are the semantics of `(::T...) where {T}` (lowered to `(::Vararg{T,N}) where {T,N}`) and `(::NTuple{N,T}) where {T,N}` identical, from the callee POV? If so, I guess we could rewrite the function defs and calls, but TBH that would not be very high on my priority list.

MikeInnes:
I think you'd need a vararg wrapper for dispatch that creates the tuple. I'm not sure where those things would sit though – agree it doesn't need to be a high priority.

PR #89, opened 2017-08-03
cudaconvert for tuples

MikeInnes:
It's pretty common to pass a variable number of arrays through a tuple, so this would be pretty useful.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -23,6 +23,7 @@ function cudaconvert(x::T) where {T}
     return x
 end
 
+cudaconvert(x::Tuple) = cudaconvert.(x)
 
 # fast lookup of global world age
 world_age() = ccall(:jl_get_tls_world_age, UInt, ())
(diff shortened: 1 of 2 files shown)


maleadt:
Good idea; add a test?

maleadt:
With [this thread](https://discourse.julialang.org/t/implementing-a-basic-genetic-algorithm-using-gpuarrays/5290/3) on Discourse, I'm wondering how far (or rather, not at all) we should take this. For example, it could be applied to structs, `cudaconvert`ing fields as well. Not sure that's sane though.

MikeInnes:
Doing it automatically for structs is unsafe, but it'd be pretty easy to define a macro that makes it opt-in (by defining the relevant CUDAnative method).

Either way, I think tuples are an important special case; they're not actually solving the same problem as structs, since the point isn't the structure but having a way to pass variable numbers of inputs.

PR #90, opened 2017-08-04
Random LLVM-related improvements

maleadt:
Could improve performance of compiler (parse less IR) and generated code (internalizing functions allows LLVM to ignore calling conventions).


maleadt:
No meaningful differences or regressions, so let's merge this.

PR #104, opened 2017-08-17
WIP: Make shuffle work for arbitrarily-sized Reals

maleadt:
Probably useful for @MikeInnes and/or @SimonDanisch (to support more element types for your arrays).

### DIFF
--- a/src/cgutils.jl
+++ b/src/cgutils.jl
@@ -132,7 +132,8 @@ end
 
 
 # create an LLVM function, given its return (LLVM) type and a vector of argument types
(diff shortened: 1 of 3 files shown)


maleadt:
Ah, this doesn't buy us much. Apparently you can't `bitcast` arbitrary aggregates, so we'll need to recurse on the individual fields...

maleadt:
Also doesn't seem to be inlining correctly.

maleadt:
Idea: use LLVM-based one for primitives, use generated function to iterate fields and shuffle those.

```
julia> type Foo
       a::Int
       b::Int
       Foo(x::Int) = new(x, x)
       end

julia> Foo(1)
Foo(1, 1)

julia> eval(Expr(:new, :Foo, :1, :2))
Foo(1, 2)
```

PR #112, opened 2017-09-06
Make CUDAnative work on Julia master again

maleadt:


maleadt:
needs LLVM tag.

maleadt:
@vchuravy FYI, this fixes the issue I mentioned to you

src/jit.jl
```diff
@@ -194,13 +194,38 @@ function add_entry!(mod::LLVM.Module, func::ANY, tt::ANY; kernel::Bool=false)
...
+                    candidate_uses = []
+                    for param in entry_params
+                        append!(candidate_uses, collect(uses(param)))
+                    end
+                    while !isempty(candidate_uses)
+                        usepair = shift!(candidate_uses)
+                        inst = user(usepair)
+
+                        if hasmetadata(inst, LLVM.MD_tbaa)
+                            metadata!(inst, LLVM.MD_tbaa)
```
vchuravy: This seems odd to me. The purpose of this according to the comment above is to clear the TBAA metadata, right?
Does `metadata!` clear the associated metadata if supplied twice?

src/jit.jl
```diff
@@ -211,8 +236,14 @@ function add_entry!(mod::LLVM.Module, func::ANY, tt::ANY; kernel::Bool=false)
             ret!(builder)
         end
 
+        # early-inline the original entry function into the wrapper
         push!(function_attributes(entry_f), EnumAttribute("alwaysinline"))
         linkage!(entry_f, LLVM.API.LLVMInternalLinkage)
+        ModulePassManager() do pm
```
vchuravy: Cool!

src/jit.jl
vchuravy: Ah `metadata!(inst, kind)` sets the associated metadata to `C_NULL` so it does clear it. Might be worth a comment or documentation in LLVM.jl?

maleadt: Yeah, I wasn't sure about a proper interface. Maybe I should wrap the metadata in a dict-like structure where `empty!` is more explicit?

maleadt: Yeah, there's a bunch of similar artificial structs in LLVM already, this is a good place for another. I'll have a look tomorrow.

maleadt: https://github.com/maleadt/LLVM.jl/pull/49/commits/224a2a8b4135c07a33a22f710bada3ff0582c75f

PR #132, opened 2017-11-13
Use LLVM for inline assembly

maleadt:


maleadt:
@vchuravy maybe we should put this, and cgutils, in LLVM.jl.

vchuravy:
I was thinking about if LLVM.jl is appropriate. But probably we can just have a submodule tools that can be used if you need it. The usage of jlctx is probably the only problem

maleadt:
Maybe an `LLVM.Interop` submodule where usage of the Julia LLVM context is implied module-wide.

maleadt:
ref https://github.com/maleadt/LLVM.jl/pull/65

PR #140, opened 2017-12-06
Fix deprecations.

maleadt:


maleadt:
`pairwise` failure due to JuliaLang/julia#24944

PR #147, opened 2018-01-16
Rework device code reflection.

maleadt:
This PR changes device code reflection to make it consistent and improve user friendliness.

Before, we extended `Base.code_*` and corresponding macro's, which was:
- inconsistent, because depending on how it was (mis)used, it could return CPU or GPU code (eg. `CUDAnative.@code_llvm` stuck in front of a proper GPU function call vs. broadcasting over a `CuArray`)
- limited, as it the macros really only could be used in front of `@cuda` calls
- annoying, because you need to specify whether you want to use Base or CUDAnative reflection

What I've tried in this PR:
- still have `code_` functions for generating code from known device method calls, but without extending Base
- remove the `@code_` macros, and replace them with `@device_code` macros that are guaranteed to always return device code

To implement those `@device_code` macros, I've added a tiny hook mechanism to CUDAnative that makes it possible to intercept compile jobs. That makes it possible to, eg., stick it in front of CPU code that calls into CUDAnative somewhere down the line, such as CuArrays.

Disadvantage of that approach: calling `@device_code` macros actually need to evaluate the RHS, so it'll have side-effects.

TODO:
- [x] extra arguments: optimize, dump_module, etc. macro kwargs are ugly
- [x] fix tests
- [x] documentation

Comments? Suggestions?

cc @denizyuret @vchuravy @MikeInnes 
ref https://github.com/denizyuret/Knet.jl/issues/198#issuecomment-349275139

### DIFF
--- a/src/jit.jl
+++ b/src/jit.jl
@@ -388,12 +388,16 @@ end
 # in which case extra metadata needs to be attached.
 function compile_function(@nospecialize(func), @nospecialize(tt), cap::VersionNumber;
                           kernel::Bool=true)
-    check_invocation(func, tt; kernel=kernel)
+    ## high-level code generation (Julia AST)
 
     sig = "$(typeof(func).name.mt.name)($(join(tt.parameters, ", ")))"
     @debug("(Re)compiling $sig for capability $cap")
 
-    # generate LLVM IR
+    check_invocation(func, tt; kernel=kernel)
+
+
+    ## low-level code generation (LLVM IR)
+
     mod, entry = irgen(func, tt)
     if kernel
         entry = wrap_entry!(mod, entry, tt)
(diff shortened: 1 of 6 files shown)


maleadt:
demo

```julia
julia> using CuArrays

julia> using CUDAnative

julia> @device_code_llvm sum(CUDAnative.sin.(cu([1,2]) .* 2))

define void @julia_broadcast_kernel_63630(%CuDeviceArray.83* nocapture readonly dereferenceable(16), [1 x [1 x i8]]* nocapture readonly dereferenceable(1), [1 x [1 x i64]]* nocapture readonly dereferenceable(8), %CuDeviceArray.83* nocapture readonly dereferenceable(16)) local_unnamed_addr #0 !dbg !4 {

}

; Function Attrs: nounwind
define void @julia_reduce_grid_63634(float, %CuDeviceArray.89* nocapture readonly dereferenceable(16), %CuDeviceArray.89* nocapture readonly dereferenceable(16), i32) local_unnamed_addr #0 !dbg !4 {

}
0.1524949f0
```

src/jit.jl
```diff
@@ -460,6 +469,10 @@ function cufunction(dev::CuDevice, @nospecialize(func), @nospecialize(tt))
         error("Device capability v$dev_cap not supported by available toolchain")
     cap = maximum(compat_caps)
 
+    if compile_hook[] != nothing
+        compile_hook[](func, tt, cap)
```
vchuravy: For similar hooks in Base I have to use `invokelatest`, maybe not necessary here since we set it to a new lambda everytime.

PR #149, opened 2018-01-27
prototype nvvm.annotations for `__launch_bounds__`

vchuravy:
I was trying to setup something similar to CUDA-C `__launch_bounds__`
mimicking how Clang does it, but I don't think the annotation is properly passed.

Any pointer/ideas how to correctly do this?

### DIFF
--- a/src/jit.jl
+++ b/src/jit.jl
@@ -152,6 +152,16 @@ function irgen(@nospecialize(func), @nospecialize(tt))
     return mod, entry
 end
 
+function addNVVMMetadata!(mod::LLVM.Module, func::LLVM.Function, name, operand)
+    #TODO: get mod from func
+    MD = metadata(mod)
+    values = Value[func, MDString(name), ConstantInt(Int32(operand))]
+    if haskey(MD, "nvvm.annotations")
+        append!(MD["nvvm.annotations"], values)
+    end
+    push!(MD, "nvvm.annotations", MDNode(values))
+end
+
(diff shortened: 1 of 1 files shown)


maleadt:
like this?
https://github.com/JuliaGPU/CUDAnative.jl/blob/d18eca306f48ccd7b0cf05cd090bbd6880e32b11/src/jit.jl#L374-L378

But we really don't now the min/max CTA size, so where do you want to get this info from?

maleadt:
Also, if we were to propagate the launch configuration to the CUDAnative compiler, we could just as well generate new LLVM IR with range metadata attached to the launch bounds intrinsics. That way, we let LLVM optimize the IR given knowledge about the launch bounds, which I presume to be more effective then only letting the PTX JIT optimize with that knowledge.

vchuravy:
Ha I thought I grep'd the repo for `nvvm`, silly me.

Does LLVM actually use that information? But yes that would sound like the best solution.

maleadt:
Yes, https://reviews.llvm.org/rL211281

maleadt:
Oh, I see where you're coming from: you want to influence the PTX JIT's register allocator (as per the hint in nvvp)?

vchuravy:
Yeah that was what I was trying to do, much less relevant now for the problem. 
I wanted to set either `--maxrregcount` or `__launch_bounds__` to allow PTXAS to optimise register usage for the know number of threads.

PR #151, opened 2018-01-28
Early-inline all functions into the entry.

maleadt:
cc @vchuravy 

Not going to merge this immediately, first want to benchmark on eg. Rodinia.


vchuravy:
I think probably we should look into using one of the other Inliners instead of using `alwaysinline` for this.

maleadt:
The basic function inliner doesn't inline the functions we want, and the C API doesn't allow to instantiate the simple inliner with custom params.

maleadt:
Tested locally.

PR #155, opened 2018-01-31
Fixes for 0.7

maleadt:


maleadt:
Only issue is the pairwise example that fails, has been doing that for a while. I need to debug that, but it doesn't hold back these improvements.

PR #167, opened 2018-03-16
windows uses the exe extension

SimonDanisch:
let me know if you want a more fine tuned solution ;)


maleadt:
Maybe `filter(lib->startswith(lib, "libjulia"), map(path->splitdir(path)[2], Libdl.dllist()))`?

Issue #163, opened 2018-02-21
Types as kernel arguments


MikeInnes:
```julia
julia> xs = cu([1,2,3])
3-element CuArray{Float32,1}:
 1.0
 2.0
 3.0

julia> Int.(xs)
ERROR: don't know how to handle argument of type DataType

julia> convert.(Int, xs)
ERROR: don't know how to handle argument of type DataType

julia> (x -> Int(x)).(xs)
3-element CuArray{Int64,1}:
 1
 2
 3
```

This is, of course, completely work-around-able, but thanks to `broadcast` it's one of those things that can easily leak a confusing error into generic Julia code.

Perhaps we can tell when compiling a kernel that an argument is unused/elided, and ignore it?

PR #168, opened 2018-03-16
Support for passing non-isbits values, as long as they are unused.

maleadt:
Implements #163. Seems kinda insane.

```
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.7.0-DEV.4473 (2018-03-06 05:55 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit 2f826b55d6* (10 days old master)
|__/                   |  x86_64-unknown-linux-gnu

julia> @device_code_llvm @cuda ((T, i)->(@cuprintf("Got %ld\n", unsafe_trunc(T, i)); return))(Int, 1.)

define void @ptxcall__6_1(%jl_value_t addrspace(10)*, double) local_unnamed_addr {
entry:
  %args.i.i = alloca %vprintf_args, align 8
  %2 = fptosi double %1 to i64, !dbg !5
  %3 = bitcast %vprintf_args* %args.i.i to i8*, !dbg !11
  call void @llvm.lifetime.start(i64 8, i8* %3), !dbg !11
  %4 = getelementptr inbounds %vprintf_args, %vprintf_args* %args.i.i, i64 0, i32 0, !dbg !11
  store i64 %2, i64* %4, align 8, !dbg !11
  %5 = call i32 @vprintf(i8* getelementptr inbounds ([9 x i8], [9 x i8]* @0, i64 0, i64 0), i8* %3), !dbg !11
  call void @llvm.lifetime.end(i64 8, i8* %3), !dbg !11
  ret void
}

julia> synchronize()
Got 1
```

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -15,14 +15,7 @@ input object `x` as-is.
 For `CuArray` objects, a corresponding `CuDeviceArray` object in global space is returned,
 which implements GPU-compatible array functionality.
 """
-function cudaconvert(x::T) where {T}
-    if T.layout == C_NULL || !Base.datatype_pointerfree(T)
-        error("don't know how to handle argument of type $T")
-    end
-
-    return x
-end
-
+cudaconvert(x) = x
 cudaconvert(x::Tuple) = cudaconvert.(x)
 
 # fast lookup of global world age
@@ -37,7 +30,7 @@ function method_age(f, tt)::UInt
 end
 
 
-isghosttype(dt) = !dt.mutable && sizeof(dt) == 0
+isghosttype(dt) = dt.isconcretetype && sizeof(dt) == 0
 
 
 """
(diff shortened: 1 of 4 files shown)


vchuravy:
Well that is quite... Interesting?

maleadt:
I guess upstream codegen could be improved to drop arguments if they are unused. But in the meantime, we can do this in CUDAnative.
In this specific case I think we can even go further, because some uses are allowed (eg. comparing values, as types are unique'd).

maleadt:
@MikeInnes Did you test this? Is this sufficient for your use cases?

PR #173, opened 2018-03-21
Fix deprecation warnings

xu3kev:
`contains(haystack, needle)` is deprecated, use `occursin(needle, haystack)` instead.
`String(io::GenericIOBuffer)` is deprecated, use `String(take!(copy(io)))` instead.
`Base.STDOUT` is deprecated, use `stdout` instead


maleadt:
Thanks!

PR #175, opened 2018-03-30
Implement device switching.

### LINKED ISSUE #159: Low-overhead device switching
We should have something like `cudaSetDevice` to easily switch devices within a process.
Ref https://github.com/JuliaGPU/CuArrays.jl/issues/52

maleadt:
Fix #159. Alternative to `CUDA_VISIBLE_DEVICES`.

cc @MikeInnes @ViralBShah

### DIFF
--- a/src/CUDAnative.jl
+++ b/src/CUDAnative.jl
@@ -40,39 +40,6 @@ include("execution.jl")
 include("reflection.jl")
 include("irvalidation.jl")
 
-const initialized = Ref{Bool}(false)
-const default_device = Ref{CuDevice}()
-const default_context = Ref{CuContext}()
-const jlctx = Ref{LLVM.Context}()
-function __init__()
-    jlctx[] = LLVM.Context(convert(LLVM.API.LLVMContextRef,
-                                   cglobal(:jl_LLVMContext, Cvoid)))
-
-    if !configured
-        @warn """CUDAnative.jl has not been successfully built, and will not work properly.
-                 Please run Pkg.build(\"CUDAnative\") and restart Julia."""
-        return
-    end
-
-    if CUDAdrv.version() != cuda_driver_version
-        error("Your set-up has changed. Please run Pkg.build(\"CUDAnative\") and restart Julia.")
-    end
-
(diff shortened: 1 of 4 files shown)


maleadt:
Also, with JuliaGPU/CUDAdrv.jl#93 this gives us the ability to do both single- and multi-process versions of multi-GPU code.

single-process:

```julia
using CUDAdrv, CUDAnative

using Test

function kernel_vadd(gpu, a, b, c)
    i = threadIdx().x + blockDim().x * ((blockIdx().x-1) + (gpu-1) * gridDim().x)
    c[i] = a[i] + b[i]

    return nothing
end

gpus = Int(length(devices()))

dims = (gpus,3,4)
a = round.(rand(Float32, dims) * 100)
b = round.(rand(Float32, dims) * 100)

# FIXME: CuArray doesn't tie in with unified memory yet
buf_a = Mem.alloc(sizeof(a), true)
Mem.upload!(buf_a, a)
d_a = CuArray{Float32,3}(dims, buf_a)
buf_b = Mem.alloc(sizeof(a), true)
Mem.upload!(buf_b, b)
d_b = CuArray{Float32,3}(dims, buf_b)
buf_c = Mem.alloc(sizeof(a), true)
d_c = CuArray{Float32,3}(dims, buf_c)

len = prod(dims)
blocks = gpus
threads = len ÷ blocks

for (gpu,dev) in enumerate(devices())
    device!(dev)
    @cuda blocks=blocks÷gpus threads=threads kernel_vadd(gpu, d_a, d_b, d_c)
end

for dev in devices()
    # NOTE: normally you'd use events and wait for them
    device!(dev)
    synchronize()
end

c = Array(d_c)
@test a+b ≈ c
```

ViralBShah:
Thank you!

maleadt:
I deviate from CUDA though in that not the first API call will create a context, but loading CUDAnative will set-up everything for device 0. This costs some memory, and takes a little time, but I don't think it matters (maybe @jekbradbury or @MikeInnes can confirm or refute this).

PR #178, opened 2018-04-06
WIP: Ignore return value of kernels

maleadt:
Split off from #176


maleadt:
OK, might be a little iffy still, but lets go forward with it as it significantly improves usability and simplifies kernel code. We can improve as we go.

PR #182, opened 2018-04-18
Use Int instead of 32-bit integers

maleadt:
Trial balloon, use Int instead of (U)Int32. We had been careful in order to preserve 32-bit precision of intrinsics, but sadly Julia is pretty bad at this and even trivial arithmetic immediately promotes to Int64.

Using Int simplifies stuff. As an additional advantage, the signedness and wider range of Int apparently allows for some better optimization by LLVM, but I'm not sure how that compares to the increased register pressure and eager `zext`s.

There's a kink though, because shuffle intrinsics require Int32 inputs, causing a inexactness check. Not sure how to deal with this.

### DIFF
--- a/examples/reduce/reduce.jl
+++ b/examples/reduce/reduce.jl
@@ -16,11 +16,11 @@ using CUDAdrv, CUDAnative
 
 # Reduce a value across a warp
 @inline function reduce_warp(op::Function, val::T)::T where {T}
-    offset = CUDAnative.warpsize() ÷ UInt32(2)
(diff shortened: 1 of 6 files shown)


MikeInnes:
> sadly Julia is pretty bad at this and even trivial arithmetic immediately promotes to Int64

Are there any Julia issues for discussion of this? Might be too late in the game to really change things here, but probably worth having the issue open either way.

maleadt:
There's https://github.com/JuliaGPU/CUDAnative.jl/issues/25 with pointers to related issues, but nothing really relevant.

PR #193, opened 2018-06-05
Enable GitLab CI/CD.

maleadt:
See https://github.com/JuliaGPU/gitlab-ci


maleadt:
So this was working properly before I moved the runner to a JuliaGPU group (group shared runner). I've contacted GitLab.

PR #197, opened 2018-06-12
Improve shared memory

maleadt:


maleadt:
Needs https://github.com/maleadt/LLVM.jl/pull/96

PR #216, opened 2018-07-04
Adapt to new inference limits

maleadt:


maleadt:
Runs into https://github.com/JuliaLang/julia/issues/27929. With that fixed, an ordinary splat might work as well, but then we'd run into the 32-arg splat limit, without a means to warn the user.

PR #217, opened 2018-07-04
Fix CI.

maleadt:


maleadt:
Breaks 0.7-beta, but who cares about that ancient release.

PR #222, opened 2018-07-27
Improve compiler performance

maleadt:


maleadt:
Works on rc2, breaks on dev due to deprecations. I'll merge this and fix deprecations over the next couple of days.

PR #228, opened 2018-08-14
Strip RefValue when invoking kernels.

maleadt:
Doesn't make much sense across the CPU/GPU barrier anyway,
and makes it possible to broadcast eg. RefValue{::CuArray}
(which isn't bits, both ::CuArray and ::RefValue).

Demo:

```julia
using CuArrays, CUDAnative

foobar(idx, A) = A[idx]
foobar.(CuArray([1]), Base.RefValue(CuArray([42])))
```

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -20,6 +20,7 @@ cudaconvert(x::Tuple) = cudaconvert.(x)
 @generated function cudaconvert(x::NamedTuple)
     Expr(:tuple, (:($f=cudaconvert(x.$f)) for f in fieldnames(x))...)
 end
+cudaconvert(x::Base.RefValue) = cudaconvert(x[])
 
 # fast lookup of global world age
 world_age() = ccall(:jl_get_tls_world_age, UInt, ())


maleadt:
> iiuc the ref there is used to opt-out of broadcasting.

Exactly so, but `Base.RefValue{CuDeviceArray}` isn't bits... so we need to strip Ref, as far as I can see.

vchuravy:
But wouldn't that mean that our broadcast kernels will treat CuArray as if it is supposed to be broadcasted?

maleadt:
No, since it's only stripped at the kernel boundary, ie. where `Broadcast.materialize!` (or whatever) calls `@cuda`.

vchuravy:
No, due to https://github.com/JuliaGPU/CuArrays.jl/blob/1252a35cf8db1126197f464eea21a0b355487f52/src/broadcast.jl#L6-L8 it is converted as part of `preprocess`

maleadt:
Oh, interesting. I didn't trigger that with my MWE that must have triggered a different path.

maleadt:
https://github.com/JuliaGPU/CuArrays.jl/commit/f3d85e04d7a795b13df74a3c1f4dc7e84efce891

Issue #227, opened 2018-08-14
Improve error message with unsupported types


maleadt:
```julia
julia> using CUDAnative

julia> kernel(a) = nothing
kernel (generic function with 1 method)

julia> @cuda kernel([1])

ERROR: type does not have a fixed size
Stacktrace:
 [1] sizeof at ./essentials.jl:410 [inlined]
 [2] isghosttype at /home/tbesard/Julia/CUDAnative/src/execution.jl:37 [inlined]
```

Issue #223, opened 2018-08-06
Improve error message when passing tracked pointer


SimonDanisch:
```Julia
julia> using CuArrays

julia> struct Point
           x::Float32
               y::Float32
               end

julia> Base.:(+)(p1::Point, p2::Point) = Point(p1.x + p2.x, p1.y + p2.y)

julia> Base.convert(::Type{Point}, x::NTuple{2, Any}) = Point(x[1], x[2])

julia> custom_types = cu(Point[(1, 2), (4, 3), (2, 2)])
3-element CuArray{Point,1}:
 Point(1.0f0, 2.0f0)
 Point(4.0f0, 3.0f0)
 Point(2.0f0, 2.0f0)

julia> result = custom_types .+ Ref(Point(2, 2))
Illegal inttoptr
          %ptrint = ptrtoint %jl_value_t addrspace(10)* %1 to i64

signal (22): SIGABRT
in expression starting at no file:0
crt_sig_handler at /home/Administrator/buildbot/worker/package_win64/build/src/home/Administrator/buildbot/worker/package_win64/build/src\signals-win.c:92
raise at C:\WINDOWS\System32\msvcrt.dll (unknown line)
abort at C:\WINDOWS\System32\msvcrt.dll (unknown line)
runOnFunction at /home/Administrator/buildbot/worker/package_win64/build/src/home/Administrator/buildbot/worker/package_win64/build/src\llvm-gc-invariant-verifier.cpp:176
_ZN4llvm13FPPassManager13runOnFunctionERNS_8FunctionE at C:\Users\sdani\AppData\Local\Julia-0.7.0-beta2\bin\LLVM.dll (unknown line)
_ZN4llvm13FPPassManager11runOnModuleERNS_6ModuleE at C:\Users\sdani\AppData\Local\Julia-0.7.0-beta2\bin\LLVM.dll (unknown line)
_ZN4llvm6legacy15PassManagerImpl3runERNS_6ModuleE at C:\Users\sdani\AppData\Local\Julia-0.7.0-beta2\bin\LLVM.dll (unknown line)
LLVMRunPassManager at C:\Users\sdani\AppData\Local\Julia-0.7.0-beta2\bin\LLVM.dll (unknown line)
macro expansion at C:\Users\sdani\.julia\packages\LLVM\FAUY\src\base.jl:18 [inlined]
[...]

PR #235, opened 2018-08-21
Improve error messages

maleadt:


maleadt:
Fix https://github.com/JuliaGPU/CUDAnative.jl/issues/227 https://github.com/JuliaGPU/CUDAnative.jl/issues/223
Requires tag of LLVM

PR #237, opened 2018-08-23
Strip invariant.load from input pointers.

maleadt:
We violate its assumption when rewriting the kernel entrypoint.


maleadt:
Candidate workaround for https://github.com/JuliaGPU/CUDAnative.jl/issues/236.

PR #241, opened 2018-08-29
Handle exceptions from throw_...

maleadt:
Apparently LLVM managed to optimize many `throw`s away, which breaks now that we don't inline everything anymore. Let's try to work around some.

### DIFF
--- a/examples/pairwise.jl
+++ b/examples/pairwise.jl
@@ -101,12 +101,14 @@ function pairwise_dist_gpu(lat::Vector{Float32}, lon::Vector{Float32})
     return Array(rowresult_gpu)
 end
 
-
-# generate reasonable data
-const n = 10000
-const lat = rand(Float32, n) .* 45
-const lon = rand(Float32, n) .* -120
-
 using Test
 
-@test pairwise_dist_cpu(lat, lon) ≈ pairwise_dist_gpu(lat, lon)
+# generate reasonable data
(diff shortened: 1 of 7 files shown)


maleadt:
So this has been pretty painful. `ptxas` is really buggy, as it turns out.

maleadt:
This passes CuArrays tests again. Summarized:we now have a couple of LLVM passes that rewrite IR to make it GPU compatible:
- replace `throw_XXX` functions (relies on the name, to be replaced with overdubbing of `Core.throw`)
- avoid `trap`/`unreachable` that cause miscompilations (#4, we should really use LLVM for this)

vchuravy:
This reminds me of `twice_return` in Julia codegen which needs special handling as well (I recall there being at least two instances).

c.f. https://github.com/JuliaLang/julia/commit/6a1cc967686b7e1da1a7f4af0c1e029d4a9d0e87

maleadt:
Similar indeed. A proper fix seems tricky though, because then you'd need LLVM to reason about the thread mask and only return when all threads are active (assuming this issue is related due to exiting from a function during divergence... but shmem seems related too). It looks like CuArrays mapreduce triggers this quite easily though, so let's hope that helps us spot any changes in upstream irgen (ie. patterns in Julia-generated IR that generate a CFG ptxas doesn't like) before running into this again.

PR #259, opened 2018-09-26
Remove at-eval from tests

maleadt:


maleadt:
Ref https://github.com/JuliaGPU/CUDAnative.jl/issues/23. Seems to be resolved by new optimizer.

maleadt:
Might have been the new scoping rules, since it doesn't work on 0.7.

Issue #120, opened 2017-09-22
Kernel objects


maleadt:
For generic code, we need to be able to query certain properties about the compiled kernel (shared memory usage, register pressure, etc) in order to determine optimal launch configuration (e.g. using the [occupancy API](https://devblogs.nvidia.com/parallelforall/cuda-pro-tip-occupancy-api-simplifies-launch-configuration/)). Making kernels into objects with metadata and methods would be a good/required first step to create an interface for this.

MikeInnes:
I think you removed the explicit `@kernel function vadd(...)` annotations, but we might ultimately need them. I think there are a few good reasons to have them:

* It's explicit that the function is (only) intended to run on the device, and can throw better errors if it's called normally
* It's easy for the macro to add necessary metadata as above (either `vadd` is a struct, or we define `metadata(::typeof(vadd)) = ...`)
* With good launch config heuristics in place, calling `vadd(...)` could transparently launch a kernel without the `@cuda` macro.

maleadt:
I've been wary of explicit `@kernel`/`@target ptx` annotations, as they discourage writing generic code. We'll need something like it though to mark GPU-specific methods when switching to Cassette for dispatch overloading.

That said, in this case I was rather thinking of making the internal compiler machinery instantiate a callable `KernelObject` for introspection purposes, and not really changing the external interface (yet).
For example:

```julia
@cuda (dims...,) foobar(foo, bar)
# expands to
kernel = CUDAnative.compile(foobar, Tuple{typeof(foo), typeof(bar)})
launch(kernel, (dims...,), foo, bar)
```

making it possible for library functions to not use the `@cuda` shorthand, but use `compile` and `launch` manually such that the library can inspect the properties of the kernel object (register usage, shared memory usage) and figure out a good launch configuration.

vchuravy:
Yeah that would seem like the right abstraction. 

@denizyuret has asked me for something similar, specifically the ability to export a kernel to a NVTPX IR file.

PR #266, opened 2018-10-19
Add a kernel object abstraction: low-level alternative to at-cuda

maleadt:
Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/120
Should give the necessary reflection tools to tackle problems like https://github.com/JuliaGPU/CuArrays.jl/issues/160

Example usage:

```julia
args = (d_a, d_b, d_c)
GC.@preserve args begin
    kernel_args = cudaconvert.(args)
    kernel = CUDAnative.compile_function(vadd, kernel_args...)
    if CUDAnative.registers(kernel) > 42 #=TODO: CUDAdrv.registers(device()) =#
        threads = 1
    else
        threads = 2
    end
    kernel(kernel_args...; threads=threads)
end
```

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -119,7 +119,7 @@ function assign_args!(code, args)
         push!(code.args, :($var = $(esc(arg))))
     end
 
-    # convert the arguments, call the compiler and launch the kernel
+    # convert the arguments, compile the function and call the kernel
     # while keeping the original arguments alive
     var_exprs = map(vars, args, splats) do var, arg, splat
          splat ? Expr(:(...), var) : var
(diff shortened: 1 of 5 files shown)


maleadt:
Review mainly intended to elicit some feedback on the current interface.

src/execution.jl
```diff
@@ -1,80 +1,70 @@
...
+Returns a named tuple.
+"""
+function memory(k::Kernel)
+    attr = attributes(k.fun)
+    local_mem = attr[CUDAdrv.FUNC_ATTRIBUTE_LOCAL_SIZE_BYTES]
+    shared_mem = attr[CUDAdrv.FUNC_ATTRIBUTE_SHARED_SIZE_BYTES]
+    constant_mem = attr[CUDAdrv.FUNC_ATTRIBUTE_CONST_SIZE_BYTES]
+    # NOTE: manual NamedTuple construction due to use of protected "local" name
+    fields = (Symbol("local"), Symbol("shared"), Symbol("constant"))
+    return NamedTuple{fields}((local_mem, shared_mem, constant_mem))
```
vchuravy: Haha nice :)

maleadt:
`cfunction` should probably accept a tuple type or tuple of types though, cfr. `cfunction`.

MikeInnes:
Cool, I like the look of this. It's perhaps a little weird to provide the kernel_args twice though?

Would it make any sense to do the cudaconvert internally, or is that going to be unsafe?

maleadt:
> It's perhaps a little weird to provide the kernel_args twice though?

Yeah, I changed it to explicitly pass the tuple type, which makes more sense given it's a call to the compiler and doesn't actually use the values.

> Would it make any sense to do the cudaconvert internally, or is that going to be unsafe?

I prefer to keep it explicit, with each function having a clear purpose. It also makes it possible to compile for e.g. `Int` and pass a `Int32` (because `cudacall` does automatic conversion cfr. `ccall`).

Issue #4, opened 2016-09-16
Shared memory + multiple function exits cause invalid results


maleadt:
Cause seems to  be an added `checkbounds`, if that even makes sense.

Repro:

``` julia
using CUDAdrv, CUDAnative

@target ptx function kernel(arr::Ptr{Int32})
    temp = @cuStaticSharedMem(Int32, (2, 1))
    tx = Int(threadIdx().x)

    if tx == 1
        for i = 1:2
            # THIS BREAKS STUFF: checkbounds(temp, i)
            Base.pointerset(temp.ptr, 1, i, 8)
        end
    end
    sync_threads()

    Base.pointerset(arr, Base.pointerref(temp.ptr, tx, 8), tx, 8)

    return nothing
end

dev = CuDevice(0)
ctx = CuContext(dev)

d_arr = CuArray(Int32, (2, 1))
@cuda (1,2) kernel(d_arr.ptr)
println(Array(d_arr))

destroy(ctx)
```

Result without `checkbounds`: `[1; 1]`. With: `[1; 0]`.

cc @cfoket

maleadt:
Not reproducible with this code anymore, but `rodinia/lud.jl` still fails with `--check-bounds=yes` probably still caused by the same underlying issue.

maleadt:
New repro:
```julia
using CUDAdrv, CUDAnative

function kernel(ptr::Ptr{Cint})
    shared = @cuStaticSharedMem(Cint, 4)

    lane = (threadIdx().x-1) % warpsize

    if lane == 0
        @boundscheck Base.checkbounds(shared, threadIdx().x)
        unsafe_store!(shared.ptr, 0, threadIdx().x)
    end

    sync_threads()
    
    val = shfl_down(Cint(32), 1, 4)
    if lane == 0
        unsafe_store!(ptr, val)
    end

    return
end

dev = CuDevice(0)
ctx = CuContext(dev)

gpu_val = CuArray(Cint, 1)
@cuda dev (1,4) kernel(gpu_val.ptr)
val = Array(gpu_val)[1]
println(val)

destroy(ctx)
```

Returns 0 with checkbounds, 32 without.

maleadt:
Managed to reduce to two sets of LLVM IR, executed using the following snippet:

```julia
using CUDAdrv, CUDAnative, LLVM

dev = CuDevice(0)
ctx = CuContext(dev)

for ir_fn in ["bug-working.ll", "bug-broken.ll"]
[...]

PR #277, opened 2018-11-05
Simplify rewriting of unreachable blocks.

maleadt:


maleadt:
@vchuravy could you try this PR to see how it affects register pressure on your application?

I've switched approaches, instead of removing `trap` and trying to replace the subsequent `unreachable` by regular control flow, I'm now calling `exit` again but while trying to hide that control flow from both LLVM and `ptxas` (by reading from a global variable). Hopefully that sufficiently confuses `ptxas` not to mess up control flow.

maleadt:
This is what it looks like:

```julia
julia> kernel(a) = (a[1]=1; nothing)
kernel (generic function with 1 method)

julia> CUDAnative.code_llvm(kernel, Tuple{CuDeviceArray{Int,2,AS.Global}})
```

```llvm
define void @julia_kernel_1({ [2 x i64], { i64 } } addrspace(11)* nocapture nonnull readonly dereferenceable(24)) local_unnamed_addr {
top:
  %1 = getelementptr { [2 x i64], { i64 } }, { [2 x i64], { i64 } } addrspace(11)* %0, i64 0, i32 0, i64 0
  %2 = getelementptr { [2 x i64], { i64 } }, { [2 x i64], { i64 } } addrspace(11)* %0, i64 0, i32 0, i64 1
  %3 = load i64, i64 addrspace(11)* %1, align 8, !tbaa !1, !invariant.load !4
  %4 = load i64, i64 addrspace(11)* %2, align 8, !tbaa !1, !invariant.load !4
  %5 = mul i64 %4, %3
  %6 = icmp slt i64 %5, 1
  br i1 %6, label %L14, label %L17

L14:                                              ; preds = %top
  call fastcc void @ptx_throw_boundserror()
  call fastcc void @opaque_exit()
  br label %opaque_unreachable

L17:                                              ; preds = %top
  %7 = getelementptr inbounds { [2 x i64], { i64 } }, { [2 x i64], { i64 } } addrspace(11)* %0, i64 0, i32 1, i32 0
  %8 = bitcast i64 addrspace(11)* %7 to i64* addrspace(11)*
  %9 = load i64*, i64* addrspace(11)* %8, align 8, !tbaa !1, !invariant.load !4
  %10 = addrspacecast i64* %9 to i64 addrspace(1)*
  store i64 1, i64 addrspace(1)* %10, align 8, !tbaa !5
  ret void

opaque_unreachable:                                ; preds = %L14, %opaque_unreachable
  br label %opaque_unreachable
}

define internal fastcc void @opaque_exit() unnamed_addr {
entry:
  %0 = icmp eq i32 1, 1
  br i1 %0, label %trap.preheader, label %loop.preheader

loop.preheader:                                   ; preds = %entry
  br label %loop

trap.preheader:                                   ; preds = %entry
  br label %trap

trap:                                             ; preds = %trap.preheader, %trap
  call void asm sideeffect "exit;", ""() #0
  br label %trap

loop:                                             ; preds = %loop.preheader, %loop
  br label %loop
}
```

`opaque_exit` to do an `exit` without `ptxas` (hopefully) realizing so, and `opaque_unreachable` performing an infinite loop to hide `unreachable` from LLVM.

vchuravy:
This is looking good. It reduced the register pressure even more (another 6 if I recall) in the kernels I was looking at.

The only snag I hit was `ptxas` complaining about:
```
┌ Info: kernel configuration
│   N = 2
│   threads = 100
│   blocks = 1
│   CUDAnative.maxthreads(kernel) = 1024
│   CUDAnative.registers(kernel) = 62
└   CUDAnative.memory(kernel) = (local = 248, shared = 0, constant = 0)
ptxas warning : Unresolved extern variable 'breaker_of_controlflow' in whole program compilation, ignoring extern qualifier
```

maleadt:
Now this code doesn't work on this branch:

```julia
using Test
using Random
using CuArrays
using GPUArrays
using CUDAnative

function main()
    @eval CUDAnative globalUnique=0
    empty!(CUDAnative.compilecache)
    Random.seed!(0)

    A = rand(1:10, 100)
    @show cpu = mapreduce(identity, +, A)

    dA = CuArray(A)
    @show gpu = mapreduce(identity, +, dA)

    @test cpu ≈ gpu
end

function Base.mapreduce(f::Function, op::Function, A::CuArray{T, N}) where {T, N}
    OT = Int
    v0 = 0

    out = CuArray{OT,1}(undef, 1)
    @cuda threads=64 reduce_kernel(f, op, v0, A, out)
    Array(out)[1]
end

function reduce_kernel(f, op, v0::T, A, result) where {T}
    tmp_local = @cuStaticSharedMem(T, 64)
    acc = v0

    # Loop sequentially over chunks of input vector
    i = threadIdx().x
    while i <= length(A)
        element = f(A[i])
        acc = op(acc, element)
        i += blockDim().x
    end

    # Perform parallel reduction
    @inbounds tmp_local[threadIdx().x] = acc
    sync_threads()

    offset = blockDim().x ÷ 2
    while offset > 0
        @inbounds if threadIdx().x <= offset
            other = tmp_local[(threadIdx().x - 1) + offset + 1]
            mine = tmp_local[threadIdx().x]
            tmp_local[threadIdx().x] = op(mine, other)
        end
        sync_threads()
        offset = offset ÷ 2
    end

    if threadIdx().x == 1
        result[blockIdx().x] = @inbounds tmp_local[1]
    end

    return
end
```

eerily similar to https://github.com/JuliaGPU/CUDAnative.jl/issues/4. even fails without the call to `exit`,  but works when not doing the unreachable transform. so either my transformation is wrong, or `ptxas` also fails when doing divergent branches...

maleadt:
Getting `Barrier error detected. Divergent thread(s) in warp` under `cuda-memcheck --tool synccheck`

maleadt:
Turns out we can leave the `exit`/`trap` and `ptxas` isn't confused as long there is no thread-divergent control flow! So I can more easily rewrite control flow since it doesn't need to be valid or non-looping -- we will have trapped the GPU.

@vchuravy could you have another look how this impacts register usage?

maleadt:
OK, works again when using `trap` instead of `exit`...

Issue #278, opened 2018-11-06
Julia 1.0.1 triggers recursion error in `throw_overflowerr_binaryop`


vchuravy:
I was putting 1.0.2-pre.0 through its paces, https://github.com/JuliaLang/julia/pull/29918, and I hit a recursion error in one of my test cases that doesn't happen on 1.0.0.
I tried it with 1.0.1 and I did hit the same issue :(, I am looking for the commit that caused it.

```
@noinline f(i) = nothing
function kernelf(elems)
           i = (blockIdx().x-1) * blockDim().x + threadIdx().x
           i > length(elems) && return nothing
           f(elems[i])
           return nothing
end

@cuda kernelf(CartesianIndices((1:1204,1:1024)))
```

```
ERROR: GPU compilation of kernelf(CartesianIndices{2,Tuple{UnitRange{Int64},UnitRange{Int64}}}) failed
KernelError: recursion is currently not supported

Try inspecting the generated code with any of the @device_code_... macros.

Stacktrace:
 [1] #IOBuffer#300 at iobuffer.jl:112
 [2] Type at none:0
 [3] print_to_string at strings/io.jl:112
 [4] #IOBuffer#299 at iobuffer.jl:91
 [5] #IOBuffer#300 at iobuffer.jl:112
 [6] print_to_string at strings/io.jl:112
 [7] throw_overflowerr_binaryop at checked.jl:157
 [8] kernelf at REPL[6]:2
```

vchuravy:
This seems like the most likely commit https://github.com/JuliaLang/julia/pull/28876/commits/ebddae967951e5c3c5ce18afa290e5996443fe65,
but the actual recursion starts here: https://github.com/JuliaLang/julia/blob/v1.0.1/base/iobuffer.jl#L92

e.g. that we are throwing a exception in `IOBuffer` that requires a the construction of an IOBuffer, due to interpolation.

```
f(i) = throw(ArgumentError("Yeah $i"))
```

```
@code_typed f(1)
CodeInfo(
1 1 ─ %1 = invoke Base.print_to_string("Yeah "::String, _2::Vararg{Any,N} where N)::String
  │   %2 = %new(Core.ArgumentError, %1)::ArgumentError              │╻ Type
  │        (Main.throw)(%2)::Union{}                                │ 
  │        $(Expr(:unreachable))::Union{}                           │ 
  └──      $(Expr(:unreachable))::Union{}                           │ 
) => Union{}
```

[...]

PR #279, opened 2018-11-07
Add a test for #278.

maleadt:
Just putting it out here when we have an upstream fix.

ref https://github.com/JuliaLang/julia/issues/29952


maleadt:
Should probably also include something that calls `print_to_string` directly, because that's what is fixed upstream (and not the idempotency issue that only gets revealed when there's recursion).
And version bound this for 1.0.2.

PR #284, opened 2018-11-13
Use CuArrays.jl instead of CUDAdrv.CuArray

maleadt:


maleadt:
Conflict: CUDAnative requires Adapt 0.4.1+, but CuArrays is bounded at Adapt <0.4.1. Needs a CuArrays tag.

maleadt:
@MikeInnes any progress on updating Flux and/or moving to proper REQUIREments? I've improved error messages so that shouldn't be a blocker anymore.

Issue #297, opened 2018-12-10
Kernel environment wrongly cached


GiggleLiu:
Here is a simple example
```julia
using CUDAnative, CuArrays
function f(x::CuVector)
    L = length(x)
    function kernel(x)
        state = (blockIdx().x-1) * blockDim().x + threadIdx().x
        if state <= L
            x[state] = 1
        end
        return
    end

    max_threads = 256
    X = min(max_threads, L)
    Y = ceil(Int, L/X)
    @cuda threads=X blocks=Y kernel(x)
    x
end

x = randn(3) |> cu
@show f(x)

x = randn(5) |> cu
@show f(x)

# OUTPUT
# f(x) = Float32[1.0, 1.0, 1.0]
# f(x) = Float32[1.0, 1.0, 1.0, 2.10655, -1.29701]
```
In the second call, the correct output should be `f(x) = Float32[1.0, 1.0, 1.0, 1.0, 1.0]`

maleadt:
Do know that this will compile an excessive amount of kernels: one for each value of `L`. Use regular parameters if you want to avoid this.

PR #298, opened 2018-12-10
Recompile kernel if captured values change

maleadt:
Fix https://github.com/JuliaGPU/CUDAnative.jl/issues/297


GiggleLiu:
Thanks for quick fix.

PR #301, opened 2018-12-16
Various improvements to docs

timholy:
I'm finally converting from old CUDArt to the newer infrastructure. I noticed a few places where I could help by improving the documentation. Now the only warnings when you include `make.jl` are related to `@ref`s to things like `InteractiveUtils.@code_llvm`, which AFAICT Documenter does not support.

The least satisfying part of this is the fact that I changed the "attachment point" for the documentation for `(::Kernel)(args...; kwargs...)`. The former choice was the most logical, but with a minute or two of experimentation I was unable to find a way to obtain that docstring from within the REPL. Consequently I moved it to the type itself, so that it can be obtained with `?CUDAnative.Kernel`.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -15,7 +15,6 @@ makedocs(
         "Library" => [
             "lib/compilation.md",
             "lib/reflection.md",
-            "lib/profiling.md",
             "Device Code" => [
                 "lib/device/intrinsics.md",
                 "lib/device/array.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,7 +2,7 @@ using Documenter, CUDAnative
 
 makedocs(
     modules = [CUDAnative],
-    format = Documenter.HTML(),
+    format = Documenter.HTML(prettyurls = get(ENV, "CI", nothing) == "true"),
     sitename = "CUDAnative.jl",
     pages = [
         "Home"    => "index.md",
--- a/docs/src/lib/compilation.md
+++ b/docs/src/lib/compilation.md
@@ -2,6 +2,15 @@
 
 ```@docs
 CUDAnative.@cuda
+CUDAnative.cufunction
+CUDAnative.Kernel
+CUDAnative.compile
 CUDAnative.cudaconvert
 CUDAnative.nearest_warpsize
+```
+
+## Devices
+
+```@docs
+CUDAnative.device!
 ```
--- a/docs/src/lib/profiling.md
+++ b/docs/src/lib/profiling.md
@@ -1,5 +1,0 @@
-# Profiling
-
-```@docs
-CUDAnative.@profile
-```
--- a/docs/src/lib/reflection.md
+++ b/docs/src/lib/reflection.md
@@ -28,3 +28,12 @@ CUDAnative.@device_code_ptx
 CUDAnative.@device_code_sass
 CUDAnative.@device_code
 ```
+
+## Version and related queries
+
+```@docs
+CUDAnative.version
+CUDAnative.maxthreads
+CUDAnative.registers
+CUDAnative.memory
+```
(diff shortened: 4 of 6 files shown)


maleadt:
Thanks! I pushed some additional fixes, there were some more inaccuracies in those sections. Docs are really a pain points of this infrastructure, let me know if there's any issues porting your code.

timholy:
:+1: As I return to the world of GPU programming and folks in my lab are also starting to learn, it occurs to me that Julia might benefit from having more "learn how to program CUDA" tutorial info written in Julia. (For people who know C, it's essentially trivial to "translate" C documentation into Julia, but for people who don't know C I imagine that may be a little more intimidating.) Is this a good idea? If so, where should such documentation be hosted?

I suspect that having translations (*would we have to get permission from the author?) of

- https://devblogs.nvidia.com/even-easier-introduction-cuda/
- possibly https://devblogs.nvidia.com/unified-memory-cuda-beginners/
- http://developer.download.nvidia.com/assets/cuda/files/reduction.pdf

might be enough to get many people started. I am not proposing doing all that myself but I could contribute to the effort, if others think it's a good idea.

maleadt:
> Is this a good idea?

I think it is. I've never really worked on them because I've never figured out a good format that is sane to commit and convenient collaborate on. @ChrisRackauckas, IIRC you recently switched notebooks over to Weave.jl for this kind of content, what are your experiences with that?
CuArrays.jl would probably be a good place to host.

I have some old content [over here](https://blog.maleadt.net/2017/07/02/juliacon17/), but that isn't up to date and not that well structured.

ChrisRackauckas:
Notebooks are terrible for keeping up to date because the diffs are unreadable, and even with Nb runner tools they are hard to CI test. The things I have translated to Weave.jl are great because you can make Weave generate not just a website, but also PDFs and Julia scripts. The libraries where I've done this are still private, but an example of the structure I've been using is found here: https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/pull/32/files . We also have it setup with a nice footer where said footer function adds details about the computer and the packages installed when running, for easier comparisons.

timholy:
Just to check, what do you see as the main weaknesses of just doing this in Documenter? I'm imagining that capturing plots would not be a major concern here, though perhaps for performance measurement that could be relevant. Are you basically hoping for something that looks fancier, or is there another motivation? (e.g., convert into a blog post, etc.)

maleadt:
> what do you see as the main weaknesses of just doing this in Documenter?

I have the mental model that Documenter.jl is good for text interspersed with snippets of code, and Weave.jl for code interspersed with text. For example, doesn't Documenter evaluate every code block in a new module, and you need an explicit set-up phase to have definitions persist? It also seems like an advantage if the code is just in an executable `.jl` file, easy to hack and improve, instead of in a `.md` file that can't easily be evaluated separately.

I can have a look at some demo later today.

timholy:
That's enough positives that I'm motivated to learn Weave.jl. I will try the "super simple" tutorial and see how it goes.

maleadt:
Had a short look: https://github.com/maleadt/tutorials/blob/master/doc/vadd.jl
https://htmlpreview.github.io/?https://github.com/maleadt/tutorials/blob/master/public/vadd.html

Thoughts?

timholy:
Definitely useful. A true beginner might need a slightly different approach, but I think we should have both. If you want to go with that I can add something as well.

maleadt:
> A true beginner might need a slightly different approach

In terms of format (ie. not using Weave.jl)? Because the content is just the `vadd` example I had lying around, definitely not intended to be the beginner's guide you are looking for.

timholy:
Yes, I've been playing with Weave.jl and am liking it.

PR #309, opened 2019-01-03
Use `local` instead of explicit call to `gensym`

KristofferC:
Arguably a bit cleaner and removes a FIXME.


maleadt:
Thanks. Not sure why I couldn't get this to work like that...
bors r+

maleadt:
My push to master killed bors.
bors r+

PR #312, opened 2019-01-04
Don't initialize CUDA when querying for the version.

maleadt:
The version happens during precompilation, and might consequently cause context
creation in the parent process to fail when eg. using process exclusive mode.


maleadt:
Possible solution for https://discourse.julialang.org/t/cuarrays-error-calling-cuarray-error-invalid-device/18654
bors try

Issue #313, opened 2019-01-04
setindex bounds checking regression on 1.1 (InvalidIRError)


maleadt:
The following works on Julia 1.0, not on 1.1:

```julia
using CuArrays

A = CuArray{Float64}(undef, 1, 1)
A[1:1, 1:1] .= 0
```

Execute with `julia --check-bounds=yes` (cfr. tests)

```
Reason: unsupported call to the Julia runtime (call to jl_gc_pool_alloc)
Stacktrace:
 [1] throw_boundserror at subarray.jl:43
 [2] checkbounds at abstractarray.jl:449
 [3] setindex! at subarray.jl:268
 [4] _setindex! at abstractarray.jl:1050
 [5] setindex! at abstractarray.jl:1020
 [6] #23 at /home/tbesard/Julia/GPUArrays/src/broadcast.jl:50
```

CUDAnative should catch calls to `throw_boundserror` and null them, which for some reason doesn't seem to be happening for this example. This is the reason tests have been failing on -dev for a while.

maleadt:
Instead of rewriting those `throw` functions we should probably try and get rid of calls to `jl_gc_pool_alloc` when the result is unused.

maleadt:
Even with https://github.com/JuliaGPU/CUDAnative.jl/pull/314, this change is more harmful since it also contains an invoke:
```
ERROR: LoadError: InvalidIRError: compiling #23(CuArrays.CuKernelState, SubArray{Float64,2,CuDeviceArray{Float64,2,CUDAnative.AS.Global},Tuple{UnitRange{Int64},UnitRange{Int64}},false}, Base.Broadcast.Broadcasted{Nothing,Tuple{Base.OneTo{Int64},Base.OneTo{Int64}},typeof(identity),Tuple{Int64}}) resulted in invalid LLVM IR
Reason: unsupported call to the Julia runtime (call to jl_box_int64)
Stacktrace:
 [1] throw_boundserror at subarray.jl:43
 [2] checkbounds at abstractarray.jl:449
 [3] setindex! at subarray.jl:268
 [4] _setindex! at abstractarray.jl:1050
 [5] setindex! at abstractarray.jl:1020
 [6] #23 at /home/tbesard/Julia/GPUArrays/src/broadcast.jl:50
Reason: unsupported call to the Julia runtime (call to jl_invoke)
Stacktrace:
 [1] throw_boundserror at subarray.jl:43
[...]

PR #314, opened 2019-01-07
Implement PTLS and GC intrinsics

maleadt:
Attempt to fix https://github.com/JuliaGPU/CUDAnative.jl/issues/313 in a better way. Makes it possible to allocate on the GPU, and thus makes the following work:

```julia
@noinline custom_throw(x) = throw(x)
function kernel(i)
    if i > 0
        custom_throw(BoundsError())
    end
    nothing
end
```

This would previously allocate, since it isn't caught by the heuristic that catches throwing functions (a common pattern on 1.0 where GC frames are put in noinline functions not to spoil the hot path).

```llvm
define void @julia_kernel_1(i64) local_unnamed_addr {
top:
  %1 = icmp slt i64 %0, 1
  br i1 %1, label %L6, label %L3

L3:                                               ; preds = %top
  %2 = call %jl_value_t* @malloc(i64 16)
  %3 = addrspacecast %jl_value_t* %2 to %jl_value_t addrspace(10)*
  %4 = bitcast %jl_value_t* %2 to %jl_value_t addrspace(10)**
  store %jl_value_t addrspace(10)* null, %jl_value_t addrspace(10)** %4, align 8, !tbaa !1
  %5 = getelementptr inbounds %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %4, i64 1
  store %jl_value_t addrspace(10)* null, %jl_value_t addrspace(10)** %5, align 8, !tbaa !1
  call fastcc void @julia_custom_throw_8317()
  call void asm sideeffect "trap;", ""() #0
  br label %L6

L6:                                               ; preds = %L3, %top
  ret void
}
```

Of course, with better analysis it should be possible to get rid of the allocation completely, but this is useful nonetheless.

WIP because it needs checked removal of the TLS getter.


maleadt:
bors try
This doesn't actually give us full allocation capabilities though, eg. disabling the `remove_throw!` pass or the `exception_arguments` hack, both of which are there to avoid certain allocations, makes Julia crash when running the CuArrays test suite. But this is a first step.

maleadt:
Looks like Julia's passes rely on intrinsics to reconstruct types.
bors try

maleadt:
`julia --check-bounds=yes -e "using CuArrays, GPUArrays; GPUArrays.TestSuite.test_mapreduce(CuArray)"` fails

maleadt:
Okay, this should be good to go. @Keno, care to comment or review? These passes probably weren't designed with an external implementation in mind though :slightly_smiling_face: 
bors try

maleadt:
cc @jonathanvdc

Keno:
Anything specific you need me to look at? I think I'm lacking context for a global review

maleadt:
> I think I'm lacking context for a global review

Yeah no worries, this isn't ready for a global review anyway. This PR is laying the groundwork to tap into some more complex Julia features that require a runtime instead of just disallowing such operations.

Specifically, I was trying to have a CUDAnative pass implement `julia.alloc_obj` with a plain malloc (baby steps). I guess ultimately it would be better to repurpose the GC lowering pass, but it's pretty CPU specific right now (TLS references, pool sized allocs, etc). Much of codegen is similar, eagerly emitting CPU-specific IR. Any thoughts on avoiding that, maybe emitting calls to lower-level intrinsics (eg. `julia.alloc_obj` -> `julia.julia.gc_alloc`) and have those separately lowered after optimization?

maleadt:
OK, with this we now support trivially allocating code:

```julia
    mutable struct PleaseAllocate
        y::Csize_t
    end

    @noinline function inner(x)
        @cuprintf("%d\n", x.y)
        return
    end

    function kernel(i)
        inner(PleaseAllocate(Csize_t(42)))
        return
    end
```

This PR also greatly improves support for the CUDAnative runtime, so it should be feasible now to provide GPU counterparts to functions like `jl_box_int`. With those, it should be possible to get rid of the hacky `replace_throw!` pass.

bors r+

PR #318, opened 2019-01-14
Implement heap-allocated boxes

maleadt:
Implements `jl_box_uint64` which makes it possible to execute the following without our hacky `replace_throw!` pass:

```julia
function foo(i, ptr)
    for Ri_base in 0:1:i
        unsafe_store!(ptr, 0)
    end
    nothing
end
```

( this constructs a lot of boxes when creating exception objects )

Should also make it possible to eg. store into `CuArray{Any}` vectors, although it's pretty unclear there who owns that memory.

Additional difficulty: type tag values change across Julia sessions, so to avoid rebuilding the CUDAnative runtime each time I emit calls to `late_jl_uint64_type` and resolve those to actual values when linking the library.

### DIFF
--- a/src/compiler/common.jl
+++ b/src/compiler/common.jl
@@ -13,9 +13,13 @@ struct CompilerContext
     blocks_per_sm::Union{Nothing,Integer}
     maxregs::Union{Nothing,Integer}
 
-    CompilerContext(f, tt, cap, kernel;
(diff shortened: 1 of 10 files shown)


maleadt:
Pattern-matching `late_..._type` function names is pretty ugly though.

maleadt:
Passes CuArrays tests :tada:
bors try

maleadt:
bors try

PR #321, opened 2019-01-18
Generate all box/unbox functions

maleadt:
Also some CI changes.
bors try


maleadt:
cc @MikeInnes, also adding reverse testing for CuArrays here.
bors r+

Issue #325, opened 2019-01-27
Reflection on invalid kernels broken


maleadt:
```
julia> using CUDAnative

julia> kernel() = throw()
kernel (generic function with 1 method)

julia> @cuda kernel()
ERROR: GPU compilation of kernel() failed
KernelError: kernel returns a value of type `Union{}`

Make sure your kernel function ends in `return`, `return nothing` or `nothing`.
If the returned value is of type `Union{}`, your Julia code probably throws an exception.
Inspect the code with `@device_code_warntype` for more details.

Stacktrace:
 [1] check_method(::CUDAnative.CompilerContext) at /home/tbesard/Julia/CUDAnative/src/compiler/validation.jl:14
 [2] #CompilerContext#54 at /home/tbesard/Julia/CUDAnative/src/compiler/common.jl:20 [inlined]
 [3] Type at /home/tbesard/Julia/CUDAnative/src/compiler/common.jl:19 [inlined]
 [4] #compile#95 at /home/tbesard/Julia/CUDAnative/src/compiler/driver.jl:36 [inlined]
 [5] compile at /home/tbesard/Julia/CUDAnative/src/compiler/driver.jl:36 [inlined]
 [6] #compile#94 at /home/tbesard/Julia/CUDAnative/src/compiler/driver.jl:18 [inlined]
 [7] compile at /home/tbesard/Julia/CUDAnative/src/compiler/driver.jl:16 [inlined]
 [8] macro expansion at /home/tbesard/Julia/CUDAnative/src/execution.jl:266 [inlined]
 [9] #cufunction#109(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(cufunction), ::typeof(kernel), ::Type{Tuple{}}) at /home/tbesard/Julia/CUDAnative/src/execution.jl:237
 [10] cufunction(::Function, ::Type) at /home/tbesard/Julia/CUDAnative/src/execution.jl:237
 [11] top-level scope at /home/tbesard/Julia/CUDAnative/src/execution.jl:205
 [12] top-level scope at gcutils.jl:87
 [13] top-level scope at /home/tbesard/Julia/CUDAnative/src/execution.jl:202

julia> @device_code_warntype @cuda kernel()
ERROR: GPU compilation of kernel() failed
KernelError: kernel returns a value of type `Union{}`

Make sure your kernel function ends in `return`, `return nothing` or `nothing`.
[...]

PR #326, opened 2019-01-27
Fix and test reflection of broken kernels.

maleadt:
Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/325


maleadt:
Hmm, coverage seems broken? It has definitely worked on 1.1 though, eg. https://gitlab.com/JuliaGPU/CUDAnative.jl/-/jobs/149248223. I've filed an issue over at https://github.com/JuliaWeb/MbedTLS.jl/issues/193

Issue #329, opened 2019-01-27
Exception traces regressed


maleadt:
MWE:

```
using CUDAnative, CuArrays

@noinline foo(a) = a[2]
bar(a) = (a[1] = foo(a); nothing)
@cuda bar(cu([1]))
```

On f272f0bdc4fe9c1f71d23642453c7c2d8e051137:

```
$ julia -g2 wip.jl
ERROR: a boundserror exception occurred during kernel execution.
 [1] checkbounds at abstractarray.jl:449
 [2] getindex at /home/tbesard/Julia/CUDAnative/src/device/array.jl:73
 [3] foo at /home/tbesard/Julia/CUDAnative/wip.jl:3
 [4] bar at /home/tbesard/Julia/CUDAnative/wip.jl:4
```

On master:
```
ERROR: a unknown exception occurred during kernel execution.
Stacktrace:
 [1] throw_boundserror at abstractarray.jl:484
```

PR #331, opened 2019-01-28
Lower throw and friends in an LLVM pass.

maleadt:
Fix https://github.com/JuliaGPU/CUDAnative.jl/issues/329
bors r+

### DIFF
--- a/src/compiler/irgen.jl
+++ b/src/compiler/irgen.jl
@@ -15,23 +15,6 @@ safe_fn(fn::String) = replace(fn, r"[^aA-zZ0-9_]"=>"_")
 safe_fn(f::Core.Function) = safe_fn(String(typeof(f).name.mt.name))
 safe_fn(f::LLVM.Function) = safe_fn(LLVM.name(f))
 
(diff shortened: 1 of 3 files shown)


src/compiler/irgen.jl
```diff
@@ -221,6 +190,75 @@ function irgen(ctx::CompilerContext)
...
+                    position!(builder, call)
+                    emit_exception!(builder, name, call)
+                    dispose(builder)
+                end
+
+                # remove the call
+                call_args = collect(operands(call))[1:end-1] # last arg is function itself
+                unsafe_delete!(LLVM.parent(call), call)
+
+                # HACK: kill the exceptions' unused arguments
```
vchuravy: DCE or ADCE is not enough?

maleadt: Sadly not, these are calls to `@nospecialize` constructors that contain allocations, or even calls to `jl_generic_apply`. LLVM doesn't touch those.

maleadt: Note that this hack used to be there in the past as well, it is now just better confined to a single pass.

Issue #265, opened 2018-10-08
Missing typevar results in japi3 function


dextorious:
When I run the following kernel (taken from a Lattice Boltzmann simulation)
```
function advance_gpu!(ρ::CuDeviceArray{T,D}, u::CuDeviceArray{SVector{D,T},D},
        δt::T, δx::T, τ::T, e::CuDeviceVector{SVector{D,T}}, w::CuDeviceVector{T}) where {D,Q,T<:AbstractFloat}
    x = (blockIdx().x-1) * blockDim().x + threadIdx().x
    y = (blockIdx().y-1) * blockDim().y + threadIdx().y
    NX, NY = size(u)
    cs = one(T) / sqrt(T(3))
    eq0 = zero(MVector{Q,T})
    # predictor step
    ρ0, u0 = zero(T), zero(SVector{D,T})
    for q ∈ 1 : Q
        i, j = Int(x-δt*e[q][1]), Int(y-δt*e[q][2])
        if     i < 1    i = NX
        elseif i > NX   i = 1   end
        if     j < 1    j = NY
        elseif j > NY   j = 1   end
        uij = u[i,j]
        uu = dot(uij, uij)
        eu = dot(e[q], uij)
        eq0[q] = ρ[i,j] * w[q] * ( one(T) + eu/cs^2 + 0.5*(eu^2 - cs*uu)/cs^4 )
        ρ0 += eq0[q]
        u0 += eq0[q] * e[q]
    end
    u0 = u0 / ρ0

    # corrector step
    ρ[x,y] = ρ0
    sumneq = zero(T)
    uu = dot(u0, u0)
    for q ∈ 1 : Q
        eu = dot(e[q], u0)
        eq1 = ρ0 * w[q] * ( one(T) + eu/cs^2 + 0.5*(eu^2 - cs*uu)/cs^4 )
        sumneq += e[q] * (eq1 - eq0[q])
    end
    u[x,y] = u0 - (one(T) - one(T)/τ) * sumneq / ρ0
    nothing
end
```

I get the following error:
```
ERROR: CUDAnative.jl encountered an unexpected internal compiler error.
Please file an issue attaching the following information, including the backtrace,
as well as a reproducible example (if possible).

InternalCompilerError: wrapper != nothing, at C:\Users\admin\.julia\packages\CUDAnative\AGfq2\src\compiler\irgen.jl:159

Compiler invocation:
 - f = advance_gpu!
[...]

Issue #103, opened 2017-08-17
CUDA 9 ptxas doesn't support sm_20


maleadt:
This shows as follows:

```
ptxas fatal   : Value 'sm_20' is not defined for option 'gpu-name'
```

I don't want to duplicate the compatibility database which now lives in CUDArt though...

maleadt:
Worse, the CUDA version as detected in CUDAnative's `build.jl` comes from the driver, and thus might disagree with the toolkit version of binaries like `ptxas`. What a mess.

maleadt:
```
for i in /home/tbesard/Work/CUDA/toolkit/*/bin/nvcc                                                                                                                            *[tb/bar_warp] 
do
$i --version | grep "Cuda compilation tools"
done
Cuda compilation tools, release 5.5, V5.5.0
Cuda compilation tools, release 6.0, V6.0.1
Cuda compilation tools, release 6.5, V6.5.12
Cuda compilation tools, release 7.0, V7.0.27
Cuda compilation tools, release 7.5, V7.5.17
Cuda compilation tools, release 8.0, V8.0.61
Cuda compilation tools, release 9.0, V9.0.102
Cuda compilation tools, release 7.5, V7.5.17
```

Looks pretty regular though, so CUDAapi should probably parse this.

maleadt:
Maybe just warn if the detected toolkit version doesn't match the driver version.

colbec:
I'm getting similar on openSUSE Leap 42.3 with CUDA 9. Build seems ok  but `Pkg.test("CUDAnative") ` fails with much output:
```
INFO: Testing using device GeForce GT 610
ptxas fatal   : Value 'sm_21' is not defined for option 'gpu-name'
...
```

in build a warning that I have multiple toolchains installed
```
INFO: Building CUDAnative
WARNING: Found multiple CUDA toolkit installations: /usr/local/cuda and /usr/local/cuda-9.0/targets/x86_64-linux
```
but in fact `/usr/local/cuda` is a symlink to `/usr/local/cuda-9.0/targets/x86_64-linux` but the build does not realize that.

maleadt:
[...]

PR #334, opened 2019-01-29
WIP: Use contextual dispatch for replacing functions

vchuravy:
On 1.1 Cassette should be performant enough for these kinds of transforms.

Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/27

@maleadt did you have a branch similar to this around?


vchuravy:
bors try

maleadt:
Yes, https://github.com/JuliaGPU/CUDAnative.jl/compare/tb/cassette
Didn't work because of plenty allocations, invokes, dispatches, etc. Is your approach different in that regard?
Also, #265.

vchuravy:
bors try

vchuravy:
As bors tells us apparently not ;)

@jrevels https://gitlab.com/JuliaGPU/CUDAnative.jl/-/jobs/153739960 is full of interesting cases.

src/context.jl
```diff
@@ -0,0 +1,44 @@
...
+# - mod, rem
+# - gamma
+# - bessel
+# - distributions
+# - unsorted
+
+using Cassette
+
+Cassette.@context CUDACtx
+const cudactx = CUDACtx()
```
jrevels: unlikely to held much, but would `Cassette.disablehooks(CUDACtx())`help anything here (if you're not using hooks, then this at least reduces our "attack surface" for debugging stuff)

vchuravy:
bors try

maleadt:
Yeah, as I feared... Let's mark this WIP then :slightly_frowning_face:

maleadt:
Same error count; inlining doesn't help.
That said, many stack traces point to `getindex` again, so maybe there's only a small number of errors remaining. I'll have another go at reducing `vadd` when I have some time.

vchuravy:
bors try

vchuravy:
Ok! The debugging session with Jarrett proved fruitful, we are down to 10ish failures :)

maleadt:
Cool! What were the changes?

jrevels:
> Cool! What were the changes?

We applied my usual Cassette issue workaround of "isolate the problematic thing and make it a contextual primitive (i.e. don't overdub into it)". The problematic thing here was the `@pure` function `datatype_align`. 

It turns out that while Cassette propagates purity to the compiler correctly, the compiler is (probably rightfully) pessimistic and just bails out on purity optimization for generated functions (i.e. `overdub`). ref https://github.com/JuliaLang/julia/pull/31012, which is my naive attempt at changing the compiler to allow this sort of thing. If that lands, we can remove the extra contextual primitive definition here.

src/context.jl
```diff
@@ -0,0 +1,52 @@
...
+          :signbit, :abs,
+          :sqrt, :cbrt,
+          :ceil, :floor,)
+    @eval function Cassette.overdub(ctx::CUDACtx, ::typeof(Base.$f), x::Union{Float32, Float64})
+        @Base._inline_meta
+        return CUDAnative.$f(x)
+    end
+end
+
+contextualize(f::F) where F = (args...) -> Cassette.overdub(cudactx, f, args...)
```
vchuravy:
```suggestion
contextualize(f::F) where F = (args...) -> (Cassette.overdub(cudactx, f, args...); return nothing)
```

We could go back to automatically returning `nothing` here

vchuravy:
bors try

vchuravy:
I think we are down to two Cassette related issues while the rest is adjustment of tests/one level of indirection missing things up.

maleadt:
```
julia> function kernel1(T, i)
           sink(i)
           return
       end
kernel1 (generic function with 1 method)

julia> @cuda kernel1(Int, 1)
ERROR: InvalidIRError: compiling #103(Type{Int64}, Int64) resulted in invalid LLVM IR
Reason: unsupported call to the Julia runtime (call to jl_f_tuple)
Stacktrace:
 [1] #103 at /home/tbesard/Julia/CUDAnative/src/context.jl:51
Reason: unsupported call to the Julia runtime (call to jl_f_getfield)
Stacktrace:
 [1] #103 at /home/tbesard/Julia/CUDAnative/src/context.jl:51
```

```
julia> inner_kwargf(foobar;foo=1, bar=2) = nothing
inner_kwargf (generic function with 1 method)

julia> @cuda (()->inner_kwargf(42;foo=1,bar=2))()
ERROR: GPU compilation of #103() failed
KernelError: kernel returns a value of type `Any`
```

Some more obscure errors as well, but these are the obvious codegen-related ones.
bors try

maleadt:
I _really_ dislike the loss of method redefinition support though, so either we need a proper fix or a hack (like emptying the CUDAnative compile cache upon every REPL execution -- but we don't have a useful REPL API for that) to support redefinitions.

vchuravy:
bors try

vchuravy:
bors try

vchuravy:
I agree that losing the ability to redefine is annoying.

Regarding:
```

**Comment by vchuravy** [comment]
```
argument count: Error During Test at /builds/JuliaGPU/CUDAnative.jl/test/device/execution.jl:440
  Got exception outside of a @test
  InvalidIRError: compiling #103(Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64, Int64) resulted in invalid LLVM IR
  Reason: unsupported call to the Julia runtime (call to jl_f__apply)
  Stacktrace:
   [1] #103 at /builds/JuliaGPU/CUDAnative.jl/src/context.jl:56
  Stacktrace:
   [1] check_ir(::CUDAnative.Com
```

Looks like the tuple limit.

bors try

**Comment by vchuravy** [comment]
Ok that reduces it down to:
- `dummy`
- 265 for Cassette, https://github.com/jrevels/Cassette.jl/issues/6
- https://github.com/JuliaGPU/CUDAnative.jl/pull/334#issuecomment-466184514
- Traces now have `overdub` in them, would be lovely to filter those out

**Comment by maleadt** [comment]
Regardless of the remaining issues, it's pretty impressive how this passes CuArrays tests!

@jrevels, do you have any ideas about a 265-workaround? It looks like a world age redesign isn't on anybody's TODO list, and a Zygote-style `refresh()` that reincludes parts of CUDAnative won't help here.

**Comment by maleadt** [comment]
The 256 issue is basically:

```julia
using Cassette
using InteractiveUtils

if true #= change me =#
    Cassette.@context CUDACtx
    const cudactx = Cassette.disablehooks(CUDACtx())

    contextualize(f::F) where F = (args...) -> Cassette.overdub(cudactx, f, args...)
else
    contextualize(f) = f
end

macro cuda(f, tt)
    quote
        f′ = contextualize($f)
        code_llvm(f′, $tt)
    end
end

kernel() = 42

@cuda kernel Tuple{}

kernel() = 100

@cuda kernel Tuple{}
```

With Cassette, this returns 42 twice, whereas the second execution should return 100. Thoughts?

jrevels:
> EDIT: hmm, this might be a bug in CUDAnative. Having a look.

Interested if you found anything, but wouldn't be surprised if this was jrevels/Cassette.jl#6

> With Cassette, this returns 42 twice, whereas the second execution should return 100. Thoughts?

Yup, this looks like jrevels/Cassette.jl#6.

> It looks like a world age redesign isn't on anybody's TODO list, and a Zygote-style `refresh()` that reincludes parts of CUDAnative won't help here.

@vtjnash and I have been talking about this for awhile, have some past attempts that hit roadblocks etc. I know it's something he's interested in working on, not sure that he's had the bandwidth to make progress on it recently though. Seems like a tough nut to crack implementation-wise 😛

vchuravy:
```
julia> @code_typed Cassette.overdub(cudactx, kernel)
CodeInfo(
1 ─     return 100
) => Int64

julia> Cassette.overdub(cudactx, kernel)
42
```

:/

maleadt:
> Seems like a tough nut to crack implementation-wise stuck_out_tongue

I had hoped it would be the irgen-equivalent of `invokelatest`, but that doesn't seem to do much. I'd think that it doesn't even need to be world-age consistent, just gimme the latest version of this code, but it's probably harder than that.

@vchuravy yeah, I'm not 100% sure what's going on. Inspecting the Method objects (`which(f, tt)`) in CUDAnative reveals that the min_age is different. I've tried a couple of things with `invokelatest`, but can't seem to get the newer code. Not sure why it works in the REPL. Doesn't when you put it in a file though.

vchuravy:
I was playing around with a pretty horrendous idea. 
Based on the observation that `overdub(ctx, f, args...) = f(args...)` works (since there is a backedge to f), what if we were to place a call to `f` into the overdubbed function, that we never call, but Julia has to assume we might.

```julia
using Cassette
function selfpass(ctx, ref)
         CI = ref.code_info
         self = GlobalRef(ref.method.module, ref.method.name)
         Cassette.insert_statements!(CI.code, CI.codelocs, 
           (x, i) -> i == 1 ?  3 : nothing,
           (x, i) -> i == 1 ? [Expr(:gotoifnot, Expr(:call, GlobalRef(Main, :unknowably_false)), i+2),
                      Expr(:call, Expr(:nooverdub, self), [Core.SlotNumber(i) for i in 2:length(CI.slotnames)]...),
                      x] : [x])
         return CI
       end
@inline function unknowably_false()
         Base.llvmcall("ret i8 0", Bool, Tuple{})
end
```

vchuravy:
Okay, this seems to work:

```julia
using Cassette
using InteractiveUtils

function selfpass(ctx, ref)
    CI = ref.code_info
    self = GlobalRef(ref.method.module, ref.method.name)
    unknowably_false = GlobalRef(Main, :unknowably_false)
    Cassette.insert_statements!(CI.code, CI.codelocs, 
      (x, i) -> i == 1 ?  4 : nothing,
      (x, i) -> i == 1 ? [
          Expr(:call, Expr(:nooverdub, unknowably_false)), 
          Expr(:gotoifnot, Core.SSAValue(i), i+3),
          Expr(:call, Expr(:nooverdub, self), [Core.SlotNumber(i) for i in 2:length(CI.slotnames)]...),
          x] : [x])
    return CI
end

@inline function unknowably_false()
    Base.llvmcall("ret i8 0", Bool, Tuple{})
end

Cassette.@context SelfCtx
const ctx = Cassette.disablehooks(SelfCtx(pass=Cassette.@pass(selfpass)))

contextualize(f::F) where F = (args...) -> Cassette.overdub(ctx, f, args...)

macro cuda(f, tt)
    quote
        f′ = contextualize($f)
        code_llvm(f′, $tt)
    end
end

kernel() = 42

@cuda kernel ()

kernel() = 10

@cuda kernel ()
```

Output:
```
➜  julia-debug ./julia -L ~/cassette.jl
WARNING: Encountered invalid lowered code for method Main.overdub(...): Core.Compiler.InvalidCodeError(kind="length(slotnames) < length(slotflags)", meta=(4, 5))

;  @ /home/vchuravy/cassette.jl:25 within `#8'
; Function Attrs: sspstrong
define i64 @"julia_#8_12233"() #0 {
top:
  ret i64 42
}
WARNING: Encountered invalid lowered code for method Main.overdub(...): Core.Compiler.InvalidCodeError(kind="length(slotnames) < length(slotflags)", meta=(4, 5))

;  @ /home/vchuravy/cassette.jl:25 within `#8'
; Function Attrs: sspstrong
define i64 @"julia_#8_12234"() #0 {
top:
  ret i64 10
}
```

@jrevels I am seeing:
```
WARNING: Encountered invalid lowered code for method Main.overdub(...): Core.Compiler.InvalidCodeError(kind="length(slotnames) < length(slotflags)", meta=(4, 5))
```

Am I forgetting updating something or am I using `insert_statements!` wrong?

maleadt:
Isn't it possible to use `jl_method_instance_add_backedge(MethodInstance, MethodInstance)`?

vchuravy:
> Isn't it possible to use `jl_method_instance_add_backedge(MethodInstance, MethodInstance)`?

Yes but also no:
1. How would cassette obtain the methodinstance to the overdub generator
2. The backedge needs to exist for each redefinition

```
using Cassette
using Cthulhu

Cassette.@context SelfCtx
const ctx = Cassette.disablehooks(SelfCtx())

kernel() = 1

mi = Cthulhu.first_method_instance(kernel, Tuple{})
overdubbed_mi = Cthulhu.first_method_instance(Cassette.overdub, Tuple{typeof(ctx), typeof(kernel)})
ccall(:jl_method_instance_add_backedge, Cvoid, (Any, Any), mi, overdubbed_mi)

@show Cassette.overdub(ctx, kernel)
kernel() = 2
@show Cassette.overdub(ctx, kernel)
kernel() = 3
@show Cassette.overdub(ctx, kernel)
```

This yields:
```
1
2
2
```

It might be possible to add an Expr node that basically is:

```
Expr(:manually_inlined, f, args...)
```

and the optimizer can then insert the corresponding backedge,m but that might be a schnapsidee.

maleadt:
Well this was a fun rebase.
@vchuravy any updates on your end? Jameson mentioned having discussed proper invalidation with you. Or should I invest time in the GPUifyLoops-like workaround?

vchuravy:
Yeah we talked about proper invalidation. That won't happen until Julia 1.3 though (at the earliest), I can update this with a couple of fixes that have accumulated for GPUifyLoops, but the 265 fix there was responsible for the rather bad compile time regression Ali experienced

maleadt:
> the 265 fix there was responsible for the rather bad compile time regression Ali experienced

It could be guarded behind `if isinteractive()`

vchuravy:
> It could be guarded behind if isinteractive()

In GPUifyLoops I have an environment variable that you can set. I pushed some of the fixes I had accumulated there.

vchuravy:
I started looking in

```
using CUDAnative; using CUDAdrv; include(joinpath(dirname(pathof(CUDAnative)), "..", "test", "array.jl"))
T = Int32
a = CuTestArray([typemax(T)])
function kernel(T, a)
           i = threadIdx().x
           @atomic a[1] = min(a[1], T(i))
           return
       end
 k = CUDAnative.contextualize(kernel)
 @code_typed k(T, cudaconvert(a))
```

and the issue looks like it is the:
```
1 ─ %1  = (getfield)(args, 2)::CuDeviceArray{Int32,1,Global}
```

which doesn't get sunk/eliminated

maleadt:
We could have a `@cuda` kwarg, `interactive=isinteractive()` to have it be compatible with the tests...

maleadt:
Implemented the above, which is much nicer IMO, but doesn't solve the problem of actually redefining the _kernel_. @vchuravy, did you fix that in another way over at GPUifyLoops?

src/context.jl
```diff
@@ -15,40 +12,50 @@
 
 using Cassette
 
-@inline function unknowably_false()
-    Base.llvmcall("ret i8 0", Bool, Tuple{})
-end
+@inline unknowably_false() = Base.llvmcall("ret i8 0", Bool, Tuple{})
+
+function generate_transform(method_redefinitions)
+    return function transform(ctx, ref)
```
vchuravy: We could also make that an option of the Ctx.

src/context.jl
```diff
@@ -15,40 +12,50 @@
...
-const InlinePass = Cassette.@pass transform
+const StaticPass = Cassette.@pass generate_transform(false)
+const InteractivePass = Cassette.@pass generate_transform(true)
 
 Cassette.@context CUDACtx
-const cudactx = Cassette.disablehooks(CUDACtx(pass = InlinePass))
+const StaticCtx = Cassette.disablehooks(CUDACtx(pass = StaticPass))
+const InteractiveCtx = Cassette.disablehooks(CUDACtx(pass = InteractivePass))
+
+@inline function contextualize(f::F, interactive) where F
```
vchuravy: We probably want `interactive` to be a `Val` to force constness.

vchuravy:
>  but doesn't solve the problem of actually redefining the kernel.

Weird, I did actually test that case: https://github.com/vchuravy/GPUifyLoops.jl/blob/951ddadfecb47b5636129d45467316aa185b32e3/test/test.jl#L40-L50
What is the code_typed of that? Is there a guarded call to the original kernel?

vchuravy:
One issue that @lcw noticed is that now all kernel are anonymous functions and that is anoying for profiling

maleadt:
Rebased and removed the 256 hacks now that we have https://github.com/JuliaLang/julia/pull/32237.
Nothing wrong with depending on a day-old alpha, right?

maleadt:
MWE:

```julia
@generated bar(::Val{align}) where {align} = :(42)
foo(i) = i+bar(Val(1))

using Cassette

function transform(ctx, ref)
    CI = ref.code_info
    noinline = any(@nospecialize(x) ->
                       Core.Compiler.isexpr(x, :meta) &&
                       x.args[1] == :noinline,
                   CI.code)
    CI.inlineable = !noinline

    CI.ssavaluetypes = length(CI.code)
    Core.Compiler.validate_code(CI)
    return CI
end
const InlinePass = Cassette.@pass transform

Cassette.@context Ctx
const ctx = Cassette.disablehooks(Ctx(pass = InlinePass))
contextualize(f::F) where F = (args...) -> Cassette.overdub(ctx, f, args...)

using InteractiveUtils
code_llvm(foo, Tuple{Int})
code_llvm(contextualize(foo), Tuple{Int})
```

Doesn't need the addition, but otherwise we get a const jlapi function. Also doesn't need the inlining pass, but otherwise the LLVM contains a call to overdub, while it now clearly shows a dynamic invocation:

```llvm
;  @ /home/tbesard/Julia/wip2.jl:2 within `foo'
define i64 @julia_foo_15956(i64) {
top:
; ┌ @ int.jl:53 within `+'
   %1 = add i64 %0, 42
; └
  ret i64 %1
}

;  @ /home/tbesard/Julia/wip2.jl:22 within `#7'
define i64 @"julia_#7_16041"(i64) {
top:
; ┌ @ /home/tbesard/Julia/wip2.jl:2 within `foo'
; │┌ @ int.jl:53 within `+'
; ││┌ @ /home/tbesard/Julia/pkg/Cassette/src/overdub.jl:465 within `_overdub_fallback'
; │││┌ @ /home/tbesard/Julia/pkg/Cassette/src/context.jl:445 within `fallback'
; ││││┌ @ /home/tbesard/Julia/pkg/Cassette/src/context.jl:447 within `call'
       %1 = add i64 %0, 42
; └└└└└
  ret i64 %1
}

```

maleadt:
265 looks broken:

```
automatic recompilation: Test Failed at /home/tbesard/Julia/pkg/CUDAnative/test/device/execution.jl:320
  Expression: (Array(arr))[] == 2
   Evaluated: 1 == 2
Stacktrace:
 [1] top-level scope at /home/tbesard/Julia/pkg/CUDAnative/test/device/execution.jl:320
 [2] top-level scope at /home/tbesard/Julia/julia-dev/build/release/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114
 [3] top-level scope at /home/tbesard/Julia/pkg/CUDAnative/test/device/execution.jl:304
 [4] top-level scope at /home/tbesard/Julia/julia-dev/build/release/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114
 [5] top-level scope at /home/tbesard/Julia/pkg/CUDAnative/test/device/execution.jl:154
 [6] top-level scope at /home/tbesard/Julia/julia-dev/build/release/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114
 [7] top-level scope at /home/tbesard/Julia/pkg/CUDAnative/test/device/execution.jl:5
non-isbits arguments: Error During Test at /home/tbesard/Julia/pkg/CUDAnative/test/device/execution.jl:324
```

Also the `jl_f_tuple` failure, and worse, the stack trace points to overdub:

```
non-isbits arguments: Error During Test at /home/tbesard/Julia/pkg/CUDAnative/test/device/execution.jl:324
  Got exception outside of a @test
  InvalidIRError: compiling #148(Type{Int64}, Int64) resulted in invalid LLVM IR
  Reason: unsupported call to the Julia runtime (call to jl_f_tuple)
  Stacktrace:
   [1] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
  Reason: unsupported call to the Julia runtime (call to jl_f_getfield)
  Stacktrace:
   [1] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
```

This one has a better on-device stack trace:
```
pow: Error During Test at /home/tbesard/Julia/pkg/CUDAnative/test/device/cuda.jl:46
  Got exception outside of a @test
  InvalidIRError: compiling #148(CuDeviceArray{Float32,0,CUDAnative.AS.Global}, Float32, Int64) resulted in invalid LLVM IR
  Reason: unsupported call to the Julia runtime (call to jl_f_tuple)
  Stacktrace:
   [1] overdub at /home/tbesard/Julia/pkg/Cassette/src/overdub.jl:508
   [2] checked_trunc_sint at boot.jl:582
   [3] toInt32 at boot.jl:619
   [4] Int32 at boot.jl:709
   [5] pow at /home/tbesard/Julia/pkg/CUDAnative/src/device/cuda/math.jl:193
   [6] pow_kernel at /home/tbesard/Julia/pkg/CUDAnative/test/device/cuda.jl:50
   [7] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
  Reason: unsupported call to the Julia runtime (call to jl_f_getfield)
  Stacktrace:
   [1] throw_inexacterror at boot.jl:560
   [2] overdub at /home/tbesard/Julia/pkg/Cassette/src/overdub.jl:0
   [3] checked_trunc_sint at boot.jl:582
   [4] toInt32 at boot.jl:619
   [5] Int32 at boot.jl:709
   [6] pow at /home/tbesard/Julia/pkg/CUDAnative/src/device/cuda/math.jl:193
   [7] pow_kernel at /home/tbesard/Julia/pkg/CUDAnative/test/device/cuda.jl:50
   [8] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
```

Some more, that fail for all their types:

```
T = Int32: Error During Test at /home/tbesard/Julia/pkg/CUDAnative/test/device/cuda.jl:780
  Got exception outside of a @test
  InvalidIRError: compiling #148(Type{Int32}, CuDeviceArray{Int32,1,CUDAnative.AS.Global}) resulted in invalid LLVM IR
  Reason: unsupported call to the Julia runtime (call to jl_f_tuple)
  Stacktrace:
   [1] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
  Reason: unsupported call to the Julia runtime (call to jl_f_getfield)
  Stacktrace:
   [1] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
```

```
T = UInt64: Error During Test at /home/tbesard/Julia/pkg/CUDAnative/test/device/cuda.jl:880
  Got exception outside of a @test
  InvalidIRError: compiling #148(Type{UInt64}, CuDeviceArray{UInt64,1,CUDAnative.AS.Global}) resulted in invalid LLVM IR
  Reason: unsupported call to the Julia runtime (call to jl_f_tuple)
  Stacktrace:
   [1] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
  Reason: unsupported call to the Julia runtime (call to jl_f_getfield)
  Stacktrace:
   [1] #148 at /home/tbesard/Julia/pkg/CUDAnative/src/context.jl:75
```


```
Test Summary:                                                  | Pass  Fail  Error  Total
CUDAnative                                                     |  348     5     32    385
  base interface                                               |                    No tests
  pointer                                                      |   20                  20
  code generation                                              |   90                  90
  code generation (relying on a device)                        |    6                   6
  execution                                                    |   70     5      1     76
    @cuda                                                      |    8     4            12
      low-level interface                                      |                    No tests
      launch configuration                                     |                    No tests
      compilation params                                       |    1                   1
      reflection                                               |    6     3             9
      shared memory                                            |                    No tests
      streams                                                  |                    No tests
      external kernels                                         |                    No tests
      calling device function                                  |                    No tests
    argument passing                                           |   27     1      1     29
      manually allocated                                       |    1                   1
      scalar through single-value array                        |    1                   1
      scalar through single-value array, using device function |    1                   1
      tuples                                                   |    1                   1
      ghost function parameters                                |    2                   2
      immutables                                               |    1                   1
      automatic recompilation                                  |    1     1             2
      non-isbits arguments                                     |                 1      1
      splatting                                                |    3                   3
      object invoke                                            |    1                   1
      closures                                                 |    1                   1
      conversions                                              |    8                   8
      argument count                                           |    4                   4
      keyword arguments                                        |                    No tests
      captured values                                          |    2                   2
    exceptions                                                 |   17                  17
    shmem divergence bug                                       |    7                   7
    dynamic parallelism                                        |   10                  10
    cooperative groups                                         |    1                   1
  pointer                                                      |   41                  41
  device arrays                                                |   20                  20
  CUDA functionality                                           |   87           31    118
    indexing                                                   |    1                   1
    math                                                       |    9            1     10
      pow                                                      |    8            1      9
    formatted output                                           |    6                   6
    @cuprint                                                   |   24                  24
    assertion                                                  |                    No tests
    shared memory                                              |   14                  14
    data movement and conversion                               |    5                   5
    clock and nanosleep                                        |                    No tests
    parallel synchronization and communication                 |   16                  16
    libcudadevrt                                               |                    No tests
    atomics (low-level)                                        |   12                  12
    atomics (high-level)                                       |                30     30
      add                                                      |                 6      6
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
        T = Float32                                            |                 1      1
        T = Float64                                            |                 1      1
      sub                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      and                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      or                                                       |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      xor                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      max                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      min                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
  examples                                                     |    6                   6
```

vchuravy:
Will look into these after my talk (especially the 265 one is disappointing). Having a reproducer with MiniCassette would be great.

maleadt:
Yeah no hurry, just reporting in that your fix works. Once merged we'll have CI at least.

maleadt:
Never mind, Pkg and I were confused about which version of Cassette we were using.
Let's try again.

maleadt:
Turns out we weren't properly contextualizing all methods. Penalty: 20 additional failures.

```
Test Summary:                                                  | Pass  Fail  Error  Total
CUDAnative                                                     |  323    24     33    380
  base interface                                               |                    No tests
  pointer                                                      |   20                  20
  code generation                                              |   63    21      1     85
    LLVM IR                                                    |   30     7            37
      basic reflection                                         |    4     1             5
      unbound typevars                                         |          1             1
      exceptions                                               |    3                   3
      sysimg                                                   |    1                   1
      child functions                                          |          1             1
      JuliaLang/julia#21121                                    |    1                   1
      kernel functions                                         |   14     2            16
        wrapper function aggregate rewriting                   |          2             2
        property_annotations                                   |   14                  14
      LLVM D32593                                              |                    No tests
      kernel names                                             |    2     2             4
      PTX TBAA                                                 |    5                   5
      tracked pointers                                         |                    No tests
      CUDAnative.jl#278                                        |                    No tests
    PTX assembly                                               |   16     8            24
      basic reflection                                         |    3                   3
      child functions                                          |          1             1
      kernel functions                                         |    6     2             8
        property_annotations                                   |    5                   5
      idempotency                                              |                    No tests
      child function reuse                                     |          2             2
      child function reuse bis                                 |                    No tests
      indirect sysimg function use                             |    2                   2
      compile for host after PTX                               |    1                   1
      LLVM intrinsics                                          |                    No tests
      kernel names                                             |    2     2             4
      exception arguments                                      |                    No tests
      GC and TLS lowering                                      |    2                   2
      float boxes                                              |          1             1
    errors                                                     |   17     6      1     24
      recursion                                                |          2             2
      base intrinsics                                          |          1      1      2
      non-isbits arguments                                     |   10                  10
      invalid LLVM IR                                          |    2                   2
      invalid LLVM IR (ccall)                                  |    2                   2
      delayed bindings                                         |    2                   2
      dynamic call (invoke)                                    |          2             2
      dynamic call (apply)                                     |    1     1             2
  code generation (relying on a device)                        |    6                   6
  execution                                                    |   72     3      1     76
    @cuda                                                      |    9     3            12
      low-level interface                                      |                    No tests
      launch configuration                                     |                    No tests
      compilation params                                       |    1                   1
      reflection                                               |    6     3             9
      shared memory                                            |                    No tests
      streams                                                  |                    No tests
      external kernels                                         |                    No tests
      calling device function                                  |                    No tests
    argument passing                                           |   28            1     29
      manually allocated                                       |    1                   1
      scalar through single-value array                        |    1                   1
      scalar through single-value array, using device function |    1                   1
      tuples                                                   |    1                   1
      ghost function parameters                                |    2                   2
      immutables                                               |    1                   1
      automatic recompilation                                  |    2                   2
      non-isbits arguments                                     |                 1      1
      splatting                                                |    3                   3
      object invoke                                            |    1                   1
      closures                                                 |    1                   1
      conversions                                              |    8                   8
      argument count                                           |    4                   4
      keyword arguments                                        |                    No tests
      captured values                                          |    2                   2
    exceptions                                                 |   17                  17
    shmem divergence bug                                       |    7                   7
    dynamic parallelism                                        |   10                  10
    cooperative groups                                         |    1                   1
  pointer                                                      |   41                  41
  device arrays                                                |   20                  20
  CUDA functionality                                           |   87           31    118
    indexing                                                   |    1                   1
    math                                                       |    9            1     10
      pow                                                      |    8            1      9
    formatted output                                           |    6                   6
    @cuprint                                                   |   24                  24
    assertion                                                  |                    No tests
    shared memory                                              |   14                  14
    data movement and conversion                               |    5                   5
    clock and nanosleep                                        |                    No tests
    parallel synchronization and communication                 |   16                  16
    libcudadevrt                                               |                    No tests
    atomics (low-level)                                        |   12                  12
    atomics (high-level)                                       |                30     30
      add                                                      |                 6      6
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
        T = Float32                                            |                 1      1
        T = Float64                                            |                 1      1
      sub                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      and                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      or                                                       |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      xor                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      max                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
      min                                                      |                 4      4
        T = Int32                                              |                 1      1
        T = Int64                                              |                 1      1
        T = UInt32                                             |                 1      1
        T = UInt64                                             |                 1      1
  examples                                                     |    6                   6
ERROR: LoadError: Some tests did not pass: 323 passed, 24 failed, 33 errored, 0 broken.
```

maleadt:
Squashed and rebased. Added a `contextualize` toggle to disable for tests that rely in naming. Maybe that mechanism would also be a way to merge this in a disabled state.

Remaining failures are almost all dynamic calls to jl_f_tuple and jl_f_getfield.

maleadt:
At least one source of those issues is the dynamic dispatch that gets introduced when passing a type. MWE:

```julia
using Cassette
Cassette.@context Noop
contextualize(f::F) where F = (args...) -> Cassette.overdub(Noop(), f, args...)

function main()
    a = [0]


    function kernel(ptr)
        unsafe_store!(ptr, 1)
        return
    end

    contextualize(kernel)(pointer(a))
    code_llvm(contextualize(kernel), Tuple{Ptr{Int}})


    function kernel(T, ptr)
        unsafe_store!(ptr, T(1))
        return
    end

    contextualize(kernel)(Int, pointer(a))
    code_llvm(contextualize(kernel), Tuple{Type{Int}, Ptr{Int}})
end
```

```llvm
define void @"julia_#34_19961"(i64) {
top:
  %1 = inttoptr i64 %0 to i64*
  store i64 1, i64* %1, align 1
  ret void
}

define void @"julia_#34_19962"(%jl_value_t addrspace(10)* nonnull, i64) {
top:
  %2 = alloca %jl_value_t addrspace(10)*, i32 2
  %gcframe = alloca %jl_value_t addrspace(10)*, i32 3
  %3 = bitcast %jl_value_t addrspace(10)** %gcframe to i8*
  call void @llvm.memset.p0i8.i32(i8* %3, i8 0, i32 24, i32 0, i1 false)
  %thread_ptr = call i8* asm "movq %fs:0, $0", "=r"()
  %ptls_i8 = getelementptr i8, i8* %thread_ptr, i64 -15712
  %ptls = bitcast i8* %ptls_i8 to %jl_value_t***
  %4 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 0
  %5 = bitcast %jl_value_t addrspace(10)** %4 to i64*
  store i64 2, i64* %5
  %6 = getelementptr %jl_value_t**, %jl_value_t*** %ptls, i32 0
  %7 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 1
  %8 = bitcast %jl_value_t addrspace(10)** %7 to %jl_value_t***
  %9 = load %jl_value_t**, %jl_value_t*** %6
  store %jl_value_t** %9, %jl_value_t*** %8
  %10 = bitcast %jl_value_t*** %6 to %jl_value_t addrspace(10)***
  store %jl_value_t addrspace(10)** %gcframe, %jl_value_t addrspace(10)*** %10
  %11 = bitcast %jl_value_t*** %ptls to i8*
  %12 = call noalias nonnull %jl_value_t addrspace(10)* @jl_gc_pool_alloc(i8* %11, i32 1400, i32 16) #1
  %13 = bitcast %jl_value_t addrspace(10)* %12 to %jl_value_t addrspace(10)* addrspace(10)*
  %14 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)* addrspace(10)* %13, i64 -1
  store %jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 140060992907120 to %jl_value_t*) to %jl_value_t addrspace(10)*), %jl_value_t addrspace(10)* addrspace(10)* %14
  %15 = bitcast %jl_value_t addrspace(10)* %12 to i64 addrspace(10)*
  store i64 %1, i64 addrspace(10)* %15, align 8
  %16 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 2
  store %jl_value_t addrspace(10)* %12, %jl_value_t addrspace(10)** %16
  %17 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %2, i32 0
  store %jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 140060989456256 to %jl_value_t*) to %jl_value_t addrspace(10)*), %jl_value_t addrspace(10)** %17
  %18 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %2, i32 1
  store %jl_value_t addrspace(10)* %12, %jl_value_t addrspace(10)** %18
  %19 = call nonnull %jl_value_t addrspace(10)* @jl_f_tuple(%jl_value_t addrspace(10)* addrspacecast (%jl_value_t* null to %jl_value_t addrspace(10)*), %jl_value_t addrspace(10)** %2, i32 2)
  %20 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 2
  store %jl_value_t addrspace(10)* %19, %jl_value_t addrspace(10)** %20
  %21 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %2, i32 0
  store %jl_value_t addrspace(10)* %19, %jl_value_t addrspace(10)** %21
  %22 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %2, i32 1
  store %jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 140060897026208 to %jl_value_t*) to %jl_value_t addrspace(10)*), %jl_value_t addrspace(10)** %22
  %23 = call nonnull %jl_value_t addrspace(10)* @jl_f_getfield(%jl_value_t addrspace(10)* addrspacecast (%jl_value_t* null to %jl_value_t addrspace(10)*), %jl_value_t addrspace(10)** %2, i32 2)
  %24 = bitcast %jl_value_t addrspace(10)* %23 to i64* addrspace(10)*
  %25 = load i64*, i64* addrspace(10)* %24, align 8
  store i64 1, i64* %25, align 1
  %26 = getelementptr %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %gcframe, i32 1
  %27 = load %jl_value_t addrspace(10)*, %jl_value_t addrspace(10)** %26
  %28 = getelementptr %jl_value_t**, %jl_value_t*** %ptls, i32 0
  %29 = bitcast %jl_value_t*** %28 to %jl_value_t addrspace(10)**
  store %jl_value_t addrspace(10)* %27, %jl_value_t addrspace(10)** %29
  ret void
}
```

PR #356, opened 2019-02-25
link against libcudadevrt

vchuravy:
I am not sure how to benchmark the increased compilation cost. This will be needed for #17.

Needs: 
- https://github.com/JuliaGPU/CUDAapi.jl/pull/65
- https://github.com/JuliaGPU/CUDAdrv.jl/pull/127

PR #357, opened 2019-02-25
Support for dynamic parallelism

### LINKED ISSUE #17: Support for Dynamic Parallelism
Dynamic parallelism support might be necessary to integrate `cuBLAS` on the device side.

Resources:
- http://users.ece.gatech.edu/~sudha/academic/class/ece8823/Lectures/Module-6-Microarchitecture/cuda-dyn-par.pdf

vchuravy:
Depends on #356 and will eventually fix #17 

There are still a couple of unanswered questions, e.g. I am not sure how to obtain the symbol name for a Julia function `f` that the linker will later resolve. Especially since we don't know that argument tuple.
I won't have time to look into this further until the weekend, but I thought I put up what I have so far!


maleadt:
Superseded by https://github.com/JuliaLang/julia/pull/31403

PR #358, opened 2019-02-27
Add Synchronisation Barriers (count, and, or)

qin-yu:
Added the following three intrinsics in `src\device\cuda_intrinsics\synchronization.jl`:
- `sync_threads_count(pred)` for LLVM IR `llvm.nvvm.barrier0.popc(i32)`;
- `sync_threads_and(pred)` for `llvm.nvvm.barrier0.and(i32)`;
- `sync_threads_or(pred)` for `llvm.nvvm.barrier0.or(i32)`.

and corresponding tests in `test\device\intrinsics.jl`.

_Although this is simple, I would like to express my special thanks of gratitude to Valentin Churavy and Tim Besard._

### DIFF
--- a/src/device/cuda_intrinsics/synchronization.jl
+++ b/src/device/cuda_intrinsics/synchronization.jl
@@ -1,6 +1,7 @@
 # Synchronization (B.6)
 
 export sync_threads, sync_warp
+export sync_threads_count, sync_threads_and, sync_threads_or
 export threadfence, threadfence_block, threadfence_system
 
 """
@@ -11,6 +12,54 @@ shared memory accesses made by these threads prior to `sync_threads()` are visib
 threads in the block.
 """
 @inline sync_threads() = ccall("llvm.nvvm.barrier0", llvmcall, Cvoid, ())
+
+"""
+    sync_threads_count(predicate::Int32)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns the number of threads for which `predicate` evaluates
+to non-zero.
+
+    sync_threads_count(predicate::Bool)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns the number of threads for which `predicate` evaluates
+to `true`.
+"""
+@inline sync_threads_count(predicate::Int32) = ccall("llvm.nvvm.barrier0.popc", llvmcall, Int32, (Int32,), predicate)
+@inline sync_threads_count(predicate::Bool) = sync_threads_count(Int32(predicate))
+
+"""
+    sync_threads_and(predicate::Int32)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns non-zero if and only if `predicate` evaluates to
+non-zero for all of them.
+
+    sync_threads_and(predicate::Bool)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns `true` if and only if `predicate` evaluates to
+`true` for all of them.
+"""
+@inline sync_threads_and(predicate::Int32) = ccall("llvm.nvvm.barrier0.and", llvmcall, Int32, (Int32,), predicate)
+@inline sync_threads_and(predicate::Bool) = ifelse(sync_threads_and(Int32(predicate)) !== Int32(0), true, false)
+
+"""
+    sync_threads_or(predicate::Int32)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns non-zero if and only if `predicate` evaluates to
+non-zero for any of them.
+
+    sync_threads_or(predicate::Int32)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns `true` if and only if `predicate` evaluates to
+`true` for any of them.
+"""
+@inline sync_threads_or(predicate::Int32) = ccall("llvm.nvvm.barrier0.or", llvmcall, Int32, (Int32,), predicate)
+@inline sync_threads_or(predicate::Bool) = ifelse(sync_threads_or(Int32(predicate)) !== Int32(0), true, false)
 
 """
     sync_warp(mask::Integer=0xffffffff)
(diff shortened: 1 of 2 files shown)


src/device/cuda_intrinsics/synchronization.jl
```diff
@@ -12,6 +13,33 @@ threads in the block.
 """
 @inline sync_threads() = ccall("llvm.nvvm.barrier0", llvmcall, Cvoid, ())
 
+"""
+    sync_threads_count(pred::Int32)
+
+is identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns the number of threads for which predicate evaluates
```
maleadt: Maybe rename the `pred` argument to `predicate` and use backticks in the docstring: "with the additional feature that it evaluate `predicate` for all threads"

maleadt:
Awesome first contribution!

I wonder if we should also have a more user friendly `sync_threads_xxx(predicate::Bool)` which can just call the underlying intrinsic passing `predicate ? Int32(1) : Int32(0)`.

maleadt:
bors try

src/device/cuda_intrinsics/synchronization.jl
qin-yu: I changed the name of argument and code-styled the corresponding keywords in the docstring. Also, the documents now starts with "Identical to ..." to match the style of the file.

qin-yu:
> Awesome first contribution!
> 
> I wonder if we should also have a more user friendly `sync_threads_xxx(predicate::Bool)` which can just call the underlying intrinsic passing `predicate ? Int32(1) : Int32(0)`.

@maleadt I felt it inconvenient to call the standard llvm ir as well!

A bool-input method has been added for each function, with appropriate user-friendly output types. I guess new methods calling old ones , e.g. `@inline sync_threads_and(predicate::Bool) = sync_threads_and(predicate ? Int32(1) : Int32(0)) > 0 ? true : false`, will not hurt performance (because, for example, `+(x::Bool, y::Bool) = Int(x) + Int(y)`)?

Tests added as well.

src/device/cuda_intrinsics/synchronization.jl
```diff
@@ -12,6 +13,54 @@ threads in the block.
...
+to non-zero.
+
+    sync_threads_count(predicate::Bool)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns the number of threads for which `predicate` evaluates
+to `true`.
+"""
+@inline sync_threads_count(predicate::Int32) = ccall("llvm.nvvm.barrier0.popc", llvmcall, Int32, (Int32,), predicate)
+@inline sync_threads_count(predicate::Bool) = sync_threads_count(predicate ? Int32(1) : Int32(0))
```
maleadt: re. friendlier output types, maybe also convert to `Int` here.

maleadt:
@vchuravy should we use `ifelse` for these simple ternaries? I'd assume llvm's if conversion is powerful enough to predicate these, but you have more experience with that.

src/device/cuda_intrinsics/synchronization.jl
vchuravy: Isn't `Int` host dependent? I would keep these as `Int32`

vchuravy:
I don't trust LLVM's if conversion. I would use `Int32(predicate)` so that we get the extend and do `% Bool` to get the truncate.

src/device/cuda_intrinsics/synchronization.jl
maleadt: It is, but we've been doing that for `threadIdx` in the assumption that we'd be able to switch that to a GPU-native type at some point. It also makes it easier for LLVM to reason about integer arithmetic, since most values (literals, indexing, etc) is dealing with CPU-native Int already.

qin-yu:
While the [CUDA Toolkit Documentation](https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#nvvm-intrin-barrier) describes `@llvm.nvvm.barrier0.and` as _"it evaluates predicate for all threads of the block and returns non-zero if and only if predicate evaluates to non-zero for all of them."_, it might not guarantee a `1` returned for all "true"/positive at some point. Is `ifelse` better than the ternary syntax?

src/device/cuda_intrinsics/synchronization.jl
vchuravy:
```suggestion
@inline sync_threads_count(predicate::Bool) = sync_threads_count(Int32(predicate))
```

src/device/cuda_intrinsics/synchronization.jl
```diff
@@ -12,6 +13,54 @@ threads in the block.
...
+non-zero for all of them.
+
+    sync_threads_and(predicate::Bool)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns `true` if and only if `predicate` evaluates to
+`true` for all of them.
+"""
+@inline sync_threads_and(predicate::Int32) = ccall("llvm.nvvm.barrier0.and", llvmcall, Int32, (Int32,), predicate)
+@inline sync_threads_and(predicate::Bool) = sync_threads_and(predicate ? Int32(1) : Int32(0)) > 0 ? true : false
```
vchuravy:
```suggestion
@inline sync_threads_and(predicate::Bool) = ifelse(sync_threads_and(Int32(predicate)) !== Int32(0), true, false)
```

src/device/cuda_intrinsics/synchronization.jl
```diff
@@ -12,6 +13,54 @@ threads in the block.
...
+non-zero for any of them.
+
+    sync_threads_or(predicate::Int32)
+
+Identical to `__syncthreads()` with the additional feature that it evaluates predicate
+for all threads of the block and returns `true` if and only if `predicate` evaluates to
+`true` for any of them.
+"""
+@inline sync_threads_or(predicate::Int32) = ccall("llvm.nvvm.barrier0.or", llvmcall, Int32, (Int32,), predicate)
+@inline sync_threads_or(predicate::Bool) = sync_threads_or(predicate ? Int32(1) : Int32(0)) > 0 ? true : false
```
vchuravy: Same as `sync_threads_and`

vchuravy:
bors try

PR #360, opened 2019-03-18
Use the primary context instead of creating our own.

maleadt:
Seems like parts of CUDA (specifically, its libraries as seen in https://github.com/JuliaGPU/CuArrays.jl/pull/294) depend on there being a proper _primary_ context.


maleadt:
bors try

PR #362, opened 2019-03-20
Dynamic parallelism

maleadt:
Builds on https://github.com/JuliaGPU/CUDAnative.jl/pull/356, requires https://github.com/JuliaLang/julia/pull/31403 and https://github.com/JuliaLang/julia/pull/31428.
Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/17, similar to https://github.com/JuliaGPU/CUDAnative.jl/pull/357 but from the other direction (first adding the necessary compiler functionality).


src/compiler/rtlib.jl
```diff
@@ -83,7 +83,7 @@ end
 
 # remove existing runtime libraries globally,
 # so any change to CUDAnative triggers recompilation
-rm(joinpath(@__DIR__, "..", "..", "deps", "runtime"); recursive=true, force=true)
+#rm(joinpath(@__DIR__, "..", "..", "deps", "runtime"); recursive=true, force=true)
```
maleadt: Should remove this.

maleadt:
Yay!

```julia

julia> function foo()
           @cuprintf("Hello, ")
           @cuda dynamic=true bar()
           CUDAnative.synchronize()
           @cuprintf("\n")
           return
       end

julia> function bar()
           @cuprintf("World!")
           return
       end

julia> @cuda foo()

julia> synchronize()
Hello, World!
```

So this approach really works. Now to polish the implementation.

maleadt:
```
using CUDAnative, CUDAdrv

function foo()
    wat(op, a, b) = (@cuprintf("Calculate: %ld\n", op(a,b)); nothing)
    @cuda dynamic=true wat(+, 1, 2)
    @cuda dynamic=true wat(-, 5, 3)
    CUDAnative.synchronize()
    return
end

@device_code_warntype @cuda foo()

synchronize()
```

```
Calculate: 3
Calculate: 2
```

Why yes obviously I expected this to work

src/compiler/optim.jl
```diff
@@ -51,9 +51,18 @@ function optimize!(ctx::CompilerContext, mod::LLVM.Module, entry::LLVM.Function)
...
             initialize!(pm)
+
+            # lower intrinsics
             add!(pm, FunctionPass("LowerGCFrame", lower_gc_frame!))
             aggressive_dce!(pm) # remove dead uses of ptls
             add!(pm, ModulePass("LowerPTLS", lower_ptls!))
+
+            # the Julia GC lowering pass also has some clean-up that is required
+            function LLVMAddLateLowerGCFramePass(PM::LLVM.API.LLVMPassManagerRef)
+                LLVM.@apicall(:LLVMExtraAddLateLowerGCFramePass,Cvoid,(LLVM.API.LLVMPassManagerRef,), PM)
```
vchuravy: Is there a PR to Base for this already? @yingboma and I were working on exposing all the Julia passes + barrier the other day but got sidetracked.

maleadt: Nicely spotted. No, I just added this. Turns out we need the pass its `CleanupIR` functionality to get rid of operation bundles; it doesn't do anything otherwise. Should I put this up, or are you going to upstream your work?

vchuravy: I was running into some linker issues, so maybe put your's up and I will add the other passes.

maleadt: https://github.com/JuliaLang/julia/pull/31428

maleadt:
OK, this should be good to go. Required _quite_ some more work than I anticipated, but it really made `@cuda dynamic=true` powerful and behave almost identical to host-side `@cuda`. Not supported: closures (due to how we drop the function type in the LLVM IR and recover it afterwards), method redefinition (there's no edge between the kernel and any other kernel function it launches).

maleadt:
@vchuravy Much simpler approach that doesn't require the actual function and tuple type in the IR: https://github.com/JuliaGPU/CUDAnative.jl/pull/362/commits/c0e11643516cf98b6d8361e1ab63775e28b11f95

PR #370, opened 2019-03-29
Support atomics

maleadt:
Still needs test, integration with CuDeviceArray (do we need an Atomic type like Base?), API matching with Base and CUDA, etc.

```julia
julia> using CUDAnative: DevicePtr

julia> code_llvm(CUDAnative.atomic_inc!, (DevicePtr{Int32,AS.Shared}, Int32))

define i32 @"julia_atomic_inc!_12382"(i64, i32) {
top:
  %2 = call i32 @llvm.nvvm.atomic.load.inc.32.p3i32(i64 %0, i32 %1)
  ret i32 %2
}

julia> code_llvm(CUDAnative.atomic_max!, (DevicePtr{UInt,AS.Generic}, UInt))

define i64 @"julia_atomic_max!_12417"(i64, i64) {
top:
    %2 = inttoptr i64 %0 to i64*
    %3 = atomicrmw umax i64* %2, i64 %1 acq_rel
  ret i64 %3
}

julia> code_llvm(CUDAnative.atomic_add!, (DevicePtr{Int,AS.Generic}, Int))

define i64 @"julia_atomic_add!_12456"(i64, i64) {
top:
    %2 = inttoptr i64 %0 to i64*
    %3 = atomicrmw add i64* %2, i64 %1 acq_rel
  ret i64 %3
}
```

### DIFF
--- a/examples/reduce/benchmark.jl
+++ b/examples/reduce/benchmark.jl
@@ -32,7 +32,7 @@ benchmark_cpu = @benchmarkable begin
 open(joinpath(@__DIR__, "reduce.jl.ptx"), "w") do f
     CUDAnative.code_ptx(f, reduce_grid, Tuple{typeof(+), CuDeviceVector{Int32,AS.Global},
                                               CuDeviceVector{Int32,AS.Global}, Int32};
-                        cap=v"6.1.0")
+                        cap=cap)
 end
 
 benchmark_gpu = @benchmarkable begin
--- a/examples/reduce/reduce.cu
+++ b/examples/reduce/reduce.cu
@@ -24,7 +24,7 @@ inline void check_code(cudaError_t code, const char *file, int line)
 __inline__ __device__
 int sumReduce_warp(int val) {
   for (int offset = warpSize/2; offset > 0; offset /= 2) 
-    val += __shfl_down(val, offset);
+    val += __shfl_down_sync(0xFFFFFFFF, val, offset);
   return val;
 }
 
--- a/examples/reduce/reduce.jl
+++ b/examples/reduce/reduce.jl
@@ -58,7 +58,6 @@ end
 # Reduce an array across a complete grid
 function reduce_grid(op::F, input::CuDeviceVector{T}, output::CuDeviceVector{T},
                      len::Integer) where {F<:Function,T}
-
     # TODO: neutral element depends on the operator (see Base's 2 and 3 argument `reduce`)
     val = zero(T)
 
--- a/src/device/array.jl
+++ b/src/device/array.jl
@@ -53,7 +53,10 @@ CuDeviceVector{T,A}(len::Integer,               p::DevicePtr{T,A}) where {T,A}  
 ## getters
 
 Base.pointer(a::CuDeviceArray) = a.ptr
+Base.pointer(a::CuDeviceArray, i::Integer) =
+    pointer(a) + (i - 1) * Base.elsize(a)
 
+Base.elsize(::Type{<:CuDeviceArray{T}}) where {T} = sizeof(T)
 Base.size(g::CuDeviceArray) = g.shape
 Base.length(g::CuDeviceArray) = prod(g.shape)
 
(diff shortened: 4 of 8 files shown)


maleadt:
Triggers https://github.com/JuliaLang/julia/issues/31624.

maleadt:
This is starting to look good.

Thoughts on a user-facing API? Methods that take a pointer don't seem very user friendly to me. But passing an additional index (`atomic_add!(a::CuArray, val, I...)`) deviates from Base.
Maybe `CUDAnative.@atomic inc!(a[I...], val)`? Could get a short-hand for common operations like add (`+=`), sub (`-=`), etc.

vchuravy:
I am not to happy with the Base interface in the first place, since it doesn't have an interface for atomically updating a location in an Array.
I always thought the right way to do that would be to say:`atomic_inc!(Ref(A, I...), val)`, or `inc!(Atomic(Ref(A, I)), val)`

maleadt:
Sure, but `Ref()` isn't GPU-compatible either (at least not its full semantics), while `@atomic` could.

vchuravy:
Oh yeah!, I don't mind `@atomic` (macros are a bit tricky to support in GPUifyLoops/code that loads CUDAnative conditionally). 

`Ref` of a `CuDeviceArray` could simply be another `CuPtr` at that offset.

maleadt:
> Oh yeah!, I don't mind `@atomic` (macros are a bit tricky to support in GPUifyLoops/code that loads CUDAnative conditionally).

Apparently @vtjnash has some thoughts about how that would look (atomic equivalents of {array,field}{ref,set}), so I came up with the following prototype of an atomic arrayset:

```julia
using CUDAnative, CUDAdrv, CuArrays
using Test

dev = device()
cap = capability(dev)

import CUDAnative: atomic_add!, atomic_sub!
# @eval CUDAnative begin
    const inplace_ops = Dict(
        :(+=) => +,
        :(-=) => -
    )

    """
        @atomic a[I] = op(a[I], val)
        @atomic a[I] ...= val
    """
    macro atomic(ex)
        # decode assignment (operator)
        if ex.head == :(=)
            ref = ex.args[1]
            rhs = ex.args[2]
            rhs.head == :call || error("right-hand side of an @atomic assignment should be a call")
            op = rhs.args[1]
            if rhs.args[2] != ref
                error("non-inplace @atomic expression should reference the same array element")
            end
            val = rhs.args[3]
        elseif haskey(inplace_ops, ex.head)
            op = inplace_ops[ex.head]
            ref = ex.args[1]
            val = ex.args[2]
        end

        # decode array expression
        if ref.head != :ref
            error("@atomic should be applied to an array reference expression")
        end
        array = ref.args[1]
        indices = Expr(:tuple, ref.args[2:end]...)

        esc(quote
            $atomic_arrayset($array, $indices, $op, $val)
        end)
    end

    # FIXME: make this respect the indexing style
    @inline atomic_arrayset(A::AbstractArray, Is::Tuple, op::Function, val) =
        atomic_arrayset(A, Base._to_linear_index(A, Is...), op, val)

    # CUDAnative.jl atomics
    for (op,impl) in [(+) => atomic_add!,
                      (-) => atomic_sub!]
        @eval @inline atomic_arrayset(A::CuDeviceArray, I::Integer, ::typeof($op), val) =
            $impl(pointer(A, I), val)
    end
# end

@testset "atomic_add" begin
    types = [Int32, Int64, UInt32, UInt64, Float32]
    cap >= v"6.0" && push!(types, Float64)

    @testset for T in types
        a = CuArray{T}([zero(T)])

        function kernel(T, a)
            @atomic a[1] += T(1)
            @atomic a[1] = a[1] + T(1)
            return
        end

        @cuda threads=1024 kernel(T, a)
        @test a[1] == T(2048)
    end
end
```

Thoughts?

vchuravy:
I prototyped a Semaphore implementation a while back in https://github.com/vchuravy/Hydrozoa.jl/blob/master/src/device/semaphore.jl might be helpful in combination with this https://stackoverflow.com/questions/2021019/implementing-a-critical-section-in-cuda and http://on-demand.gputechconf.com/gtc/2013/presentations/S3101-Atomic-Memory-Operations.pdf

maleadt:
OK, interesting. I don't have an actual use case for this functionality yet, so please comment if you see any issues. I've added a note that the `@atomic` macro is experimental, but the `atomic_...!` methods should be fine.

PR #373, opened 2019-04-03
Support for resetting the device.

maleadt:
~TODO: make the compilecache `device => ...` (we only really support a single context in CUDAnative) and wipe that.~


maleadt:
cc @jonathanvdc
bors r+

jonathanvdc:
Thanks! This'll come in handy.

PR #372, opened 2019-04-02
Grid Synchronization with Cooperative Groups

qin-yu:
Introduced two functions `this_grid()` and `sync_grid()` to implement synchronisation of all blocks within a grid. One can now pass a keyword arguement `cg::Bool=false` to `@cuda` and launch cooperative kernel (implemented in CUDAdrv, see [my other pull request](https://github.com/JuliaGPU/CUDAdrv.jl/pull/134) in JuliaGPU/CUDAdrv.jl)

PR #377, opened 2019-04-09
WIP: Generalise Grid Synchronisation to Multi-Grid/Multi-Device

qin-yu:
Extracted `_get_handle()` and `_sync_handle()` for both grid group and multi-grid group. Also, added `cudaCGScope` to the same file as it is only used here. However, this cannot be used without a `cuLaunchCooperativeKernelMultiDevice()` CUDA API that will allow a single host thread to launch a kernel across multiple devices. This should be added to CUDAdrv.jl.

- [ ]  Add multi-device cooperative kernel launch to CUDAdrv
- [x]  Add `this_multi_grid()` and `sync_multi_grid()` to CUDAnative
- [ ]  Add tests for multi-device to CUDAnative

(Grid synchronisation was done in [Cooperative Kernel Launch #134](https://github.com/JuliaGPU/CUDAdrv.jl/pull/134) and [Grid Synchronization with Cooperative Groups #372](https://github.com/JuliaGPU/CUDAnative.jl/pull/372))

### DIFF
--- a/src/device/cuda/cooperative_groups.jl
+++ b/src/device/cuda/cooperative_groups.jl
@@ -1,6 +1,26 @@
 # C. Cooperative Groups
 
-export this_grid, sync_grid
+export this_grid, sync_grid,
+       this_multi_grid, sync_multi_grid
+
+@enum cudaCGScope::Cuint begin
+     CGScopeInvalid   = 0  # Invalid cooperative group scope
+     CGScopeGrid           # Scope represented by a grid_group
+     CGScopeMultiGrid      # Scope represented by a multi_grid_group
+end
+
+
+# Intrinsic to get a handle for both grid groups and multi-grid groups:
+if VERSION >= v"1.2.0-DEV.512"
+    @inline cg_intrinsic_handle(cooperative_scope::cudaCGScope) =
+        ccall("extern cudaCGGetIntrinsicHandle", llvmcall, Culonglong, (cudaCGScope,), cooperative_scope)
+else
+    @eval @inline cg_intrinsic_handle(cooperative_scope::cudaCGScope) = Base.llvmcall(
+        $("declare i64 @cudaCGGetIntrinsicHandle(i32)",
+          "%rv = call i64 @cudaCGGetIntrinsicHandle(i32 %0)
+           ret i64 %rv"), Culonglong,
+        Tuple{cudaCGScope}, cooperative_scope)
+end
 
 """
     this_grid()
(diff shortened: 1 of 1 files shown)


maleadt:
Your `llvmcall` invocations should eval the argument tuple, eg. `@eval Base.llvmcall($(decl, ir), ...)`, not the inner string (which you don't seem to be interpolating into anyway). See https://github.com/JuliaGPU/CUDAnative.jl/pull/379

src/device/cuda/cooperative_groups.jl
```diff
@@ -1,42 +1,71 @@
...
 
-Returns a `grid_handle` of the grid group this thread belongs to. Only available if a cooperative
-kernel is launched.
-"""
-this_grid
 
+# Both grid groups and multi-grid groups use the same intrinsic to get handle:
 if VERSION >= v"1.2.0-DEV.512"
-    @inline this_grid() = ccall("extern cudaCGGetIntrinsicHandle", llvmcall, Culonglong, (Cuint,), UInt32(1))
+    @inline _get_handle(cooperative_scope::cudaCGScope) =
```
maleadt: maybe rename this to `cg_intrinsic_handle`; without namespacing this could clash or get confused with other handle functions

src/device/cuda/cooperative_groups.jl
```diff
@@ -1,42 +1,71 @@
 # C. Cooperative Groups
 
-export this_grid, sync_grid
+export this_grid, sync_grid,
+       this_multi_grid, sync_multi_grid
 
-"""
-    this_grid()
+@enum cudaCGScope::Cuint begin
+      cudaCGScopeInvalid   = 0  # Invalid cooperative group scope
```
maleadt: drop the redundant `cuda` on the values?

qin-yu:
> Your `llvmcall` invocations should eval the argument tuple, eg. `@eval Base.llvmcall($(decl, ir), ...)`, not the inner string (which you don't seem to be interpolating into anyway).
> See #379 (that PR is probably also the source of conflict with `master` here).

Oh, yes, I made this PR before I read how you saved my `grid_sync()` in your bug fix PR. Changing it now.

src/device/cuda/cooperative_groups.jl
qin-yu: Sure. (didn't learn from the CUDAapi PR 😂)

qin-yu:
I didn't need it anymore so I stopped updating. Now I found that this was implemented in CUDA.jl(https://github.com/JuliaGPU/CUDA.jl)!

PR #378, opened 2019-04-09
explain why an argument is not isbits

vchuravy:
@dpsanders mentioned recently that it is sometimes non-trivial to see why a certain argument is not isbits.
This tries to recursively walk the tree to see if we can give a better explanation.

### DIFF
--- a/test/codegen.jl
+++ b/test/codegen.jl
@@ -469,12 +469,27 @@ end
 
 # some validation happens in `compile`
 
+@eval Main begin
+struct CleverType{T}
+    x::T
+end
+Base.unsafe_trunc(::Type{Int}, x::CleverType) = unsafe_trunc(Int, x.x)
+end
+
 @testset "non-isbits arguments" begin
     foobar(i) = (sink(unsafe_trunc(Int,i)); return)
 
     @test_throws_message(CUDAnative.KernelError,
                          CUDAnative.codegen(:ptx, CUDAnative.CompilerJob(foobar, Tuple{BigInt}, cap, true))) do msg
         occursin("passing and using non-bitstype argument", msg) &&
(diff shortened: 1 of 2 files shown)


test/codegen.jl
```diff
@@ -469,6 +469,14 @@ end
 
 # some validation happens in `compile`
 
+# This is required for 1.0 -- no type definition in local-scope
```
maleadt: We do that plenty of times, doesn't really need a specific comment here.

maleadt:
Cool! Got some example output? I wonder if it proves useful in the case of broadcasting, where you end up with non-isbits Broadcastable objects when doing something wrong.

vchuravy:
David send me this one (the old header would be prepended):
```
julia> print(msg)
  .variables is of type Array{Symbol,1} which is not isbits.
  .forward is of type IntervalConstraintProgramming.GeneratedFunction{getfield(IntervalConstraintProgramming, Symbol("##19#20"))} which is not isbits.
    .code is of type Expr which is not isbits.
      .head is of type Symbol which is not isbits.
      .args is of type Array{Any,1} which is not isbits.
  .backward is of type IntervalConstraintProgramming.GeneratedFunction{getfield(IntervalConstraintProgramming, Symbol("##21#22"))} which is not isbits.
    .code is of type Expr which is not isbits.
      .head is of type Symbol which is not isbits.
      .args is of type Array{Any,1} which is not isbits.
  .expression is of type Operation which is not isbits.
    .args is of type Array{Expression,1} which is not isbits.
```

dpsanders:
That was a type that I was trying to broadcast. Note. None of the non-isbits parts are actually used in the broadcast, but I naively didn't realise that the whole object is sent over anyway!

PR #394, opened 2019-04-24
pow function for exponet is Int64

chengchingwen:
ref: https://github.com/JuliaGPU/CuArrays.jl/pull/325


chengchingwen:
@vchuravy Let's all right :)
I'm not familiar with testing CUDAnative, and I didn't find the test for the `CUDAnative.pow` in the test folder. Could tell me what/how should I do?

vchuravy:
I think this is the right place https://github.com/JuliaGPU/CUDAnative.jl/blob/59166089cda2fc579023b2865fdb7b59fc983f0e/test/device/cuda.jl#L33

chengchingwen:
I add a few test for the `pow` function. Is these ok?

vchuravy:
bors try

maleadt:
Thanks, great contribution! Though I've been purposefully not adding such extra methods, since I had hoped to keep these wrappers low-level and integrate them with eg. `Base.^` as soon as we have Cassette. But it can't hurt to add a couple for convenience, since Cassette seems to take a while.

PR #400, opened 2019-05-03
Make the kernel transform mandatory

vchuravy:
The unoptimised llvm module is currently the one produced by Julia, I find it quite helpful to also look at the one that we produce after turning it into a gpu kernel.

### DIFF
--- a/src/compiler/driver.jl
+++ b/src/compiler/driver.jl
@@ -111,6 +111,10 @@ function codegen(target::Symbol, job::CompilerJob;
             end
         end
 
+        if job.kernel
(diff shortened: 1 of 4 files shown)


maleadt:
Maybe the kernel transformation ought to be mandatory, like the other transformations in `irgen`. After all, CUDAnative expects the transformation to happen as it changes the calling convention. So then we'd still just have `optimize=true|false` (although an optimization level could be useful, but I'd rather encode is at `optimize=:none|:minimal|:full` then).

vchuravy:
> Maybe the kernel transformation ought to be mandatory

Yeah that seems like the better solution.

vchuravy:
Okay `optimize!` no longer takes the `optimize` flag.

Issue #398, opened 2019-04-29
performance gap with clang


lcw:
@tcew, @jkozdon, @vchuravy, and I have been working getting similar performance in Julia for a prototypically kernel for the types of codes we write.  You may find the example code [here][0].

 In doing this we have compared Clang version 6.0.1 that comes with julia against Julia/CUDAnative version
```
❯ julia --project
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.2.0-pre.25 (2019-04-20)
 _/ |\__'_|_|_|\__'_|  |  backports-release-1.2/9501bd2a2e (fork: 26 commits, 19 days)
|__/                   |

(volumerhs-small) pkg> status
    Status `~/julia/code/Heptapus.jl/examples/volumerhs-small/Project.toml`
  [c5f51814] CUDAdrv v3.0.0
  [be33ccc6] CUDAnative v2.1.0
  [3a865a2d] CuArrays v1.0.2
  [ba82f77b] GPUifyLoops v0.2.3
  [90137ffa] StaticArrays v0.10.3
  [9a3f8284] Random
```
 and see a performance difference.  The hope is we can close this performance gap since they are both using the same version of LLVM under the hood.

We have created the [cuda][1] and [Julia][2] kernels to be as close as possible but it seems as though the Julia version is spilling registers (the Julia kernel is using 255 registers and the clang kernel is using 80).  When I run the codes with `nvprof` I see a 5x speed difference.

```
❯ nvprof --print-gpu-trace julia --project=. volumerhs.jl
==34950== NVPROF is profiling process 34950, command: julia --project=. volumerhs.jl
==34950== Profiling application: julia --project=. volumerhs.jl
==34950== Profiling result:
   Start  Duration            Grid Size      Block Size     Regs*    SSMem*    DSMem*      Size  Throughput  SrcMemType  DstMemType           Device   Context    Stream  Name
17.8386s  924.86us                    -               -         -         -         -  9.5367MB  10.070GB/s    Pageable      Device  Tesla V100-SXM2         1         7  [CUDA memcpy HtoD]
[...]

PR #401, opened 2019-05-03
nvdisasm doesn't reliably work on CUDA >= 9.1 for LLVM 6.0.1

### LINKED ISSUE #396: `nvdisasm` needs version CUDA 9.0 use cuobjdump instead if newer

vchuravy:
fixes #396

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -142,6 +142,7 @@ function main()
     config[:libcudadevrt] = find_libcudadevrt(toolkit_dirs)
     config[:libcudadevrt] === nothing && build_error("Available CUDA toolchain does not provide libcudadevrt")
     ## optional
+    config[:cuobjdump] = find_cuda_binary("cuobjdump", toolkit_dirs)
     config[:nvdisasm] = find_cuda_binary("nvdisasm", toolkit_dirs)
     config[:ptxas] = find_cuda_binary("ptxas", toolkit_dirs)
 
--- a/src/reflection.jl
+++ b/src/reflection.jl
@@ -103,8 +103,18 @@ function code_sass(io::IO, job::CompilerJob)
     if !job.kernel
         error("Can only generate SASS code for kernel functions")
     end
-    if ptxas === nothing || nvdisasm === nothing
-        error("Your CUDA installation does not provide ptxas or nvdisasm, both of which are required for code_sass")
+
+    # nvdisams on CUDA >= 9.1 fails on the binaries produced by LLVM
+    if nvdisasm === nothing || Base.libllvm_version <= v"6.0.1" && cuda_driver_version >= v"9.1.0"
+        use_nvdisasm = false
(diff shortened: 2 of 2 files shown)


maleadt:
Can you elaborate? I've been successfully (AFAIK) using `nvdisasm`/`code_sass` on CUDA 10.1...

vchuravy:
One the kernels in #398 @lcw observed:

```
vchuravy@cyclops ~/H/e/volumerhs-small> ~/julia/julia --project=. volumerhs.jl 
[ Info: Building the CUDAnative run-time library for your sm_70 device, this might take a while...
nvdisasm error   : Untrackable indirect branch (needs a SIBCALL annotation?)
nvdisasm error   : Untrackable indirect branch (needs a SIBCALL annotation?)
nvdisasm error   : Untrackable indirect branch (needs a SIBCALL annotation?)
nvdisasm error   : Untrackable indirect branch (needs a SIBCALL annotation?)
nvdisasm error   : Untrackable indirect branch (needs a SIBCALL annotation?)
ERROR: Error while loading expression starting at /home/vchuravy/Heptapus.jl/examples/volumerhs-small/volumerhs.jl:243
caused by [exception 1]
failed process: Process(`/usr/bin/nvdisasm --print-code --print-line-info /tmp/jl_BP6M9o`, ProcessExited(1)) [1]
```
If `ptxas` from 9.0 is disassembly is successful and Clang 6.0.1 only support 9.0
I hadn't had a chance to test with CUDA 10

maleadt:
> If `ptxas` from 9.0 is disassembly is successful and Clang 6.0.1 only support 9.0

I can't parse that...

Could it be that you're using a CUDA toolkit (from which we pick up `nvdisasm`) with a CUDA version that is below the CUDA version as reported by the driver? Because, and this is just a guess, `nvdisasm` is 22MB so it might include a copy of the compiler and its disassembler while `cuobjdump` is only 182K and looks like it might be reusing the compiler infrastructure as part of `libcuda` (both don't dynamically link against any CUDA library).

maleadt:
Good to go?

maleadt:
So it looks like `cuobjdump` just uses `nvdisasm`:

```
shell> cuobjdump --dump-sass /tmp/test.cubin

        code for sm_75
                Function : ptxcall_identity_4
nvdisasm fatal   : Value 'SM75' is not defined for option 'binary'
```

maleadt:
We're now dumping the CUBIN as generated by the CUDA driver, so the behavior of this might have changed. Furthermore, CUPTI documents that we should use nvdisasm. There's possibility for a mismatch though, if you're using a CUDA toolkit that's much older than the CUDA capabilities of your driver. Either way, I don't think we should switch to cuobjdump right now (esp. since we can't ship either nvdisasm or cuobjdump when switching to BB, so I'd rather only use one of those).

PR #403, opened 2019-05-11
[skip ci] Add CITATION.bib

giordano:
Compared to the original BibTeX entry, I also added the information about the preprint on arXiv


maleadt:
Thanks!

PR #404, opened 2019-05-14
improve backtraces when used with Cassette/GPUifyLoops

vchuravy:


trontrytel:
@lcw @jkozdon and I tried it with updated GPUifyloops and we still got anonymous in nvprof

vchuravy:
This only improves the errors reported by CUDAnative, I am still thinking about how we can the profiling information properly in Cassette.

maleadt:
>  I am still thinking about how we can the profiling information properly in Cassette

Embed a custom `LineInfoNode` node in the IR for `overdub`?

vchuravy:
> Embed a custom LineInfoNode node in the IR for overdub?

I haven't tried that, but as far as I know those have been tugged away into `CodeInfo.codelocs` and `CodeInfo.linetable`, but I will mess around with that next week.

vchuravy:
I can't reproduce this anymore :/

vchuravy:
Right! The issue is when we hit an InvalidIR error, the maintained stack by the verifier only has `overdub` and doesn't reconstruct the inline frames

src/compiler/irgen.jl
```diff
@@ -20,6 +20,14 @@ function backtrace(job::CompilerJob, call_stack::Vector{Core.MethodInstance})
     bt = StackTraces.StackFrame[]
     for method_instance in call_stack
         method = method_instance.def
+        if method.name === :overdub && isdefined(method, :generator)
+            # The inline frames are maintained by the dwarf based backtrace, but here we only have the
+            # calls to overdub directly, the backtrace therefore is collapsed and we have to
+            # lookup the overdubbed function, but only if we likely are using the generated variant.
+            actual_sig = Tuple{method_instance.specTypes.parameters[3:end]...}
+            m = ccall(:jl_gf_invoke_lookup, Any, (Any, UInt), tt, typemax(UInt))
```
vchuravy: Undefined variable `tt`...

PR #414, opened 2019-06-03
change name of entry

vchuravy:


vchuravy:
Actually there are a couple more fixes I should add here

src/execution.jl
```diff
@@ -338,12 +338,13 @@ The following keyword arguments are supported:
   multiprocessor
 - `maxregs`: the maximum number of registers to be allocated to a single thread (only
   supported on LLVM 4.0+)
+- `name`: the name of the kernel entry function, if non provided this is `nameof(f)`
```
maleadt:
```suggestion
- `name`: override the name that the kernel will have in the generated code
```

PR #432, opened 2019-07-05
add abs for complex

Roger-luo:
I'm not sure if this is the right place to add it. But I think this is not there in the kernel.

### DIFF
--- a/src/device/cuda/math.jl
+++ b/src/device/cuda/math.jl
@@ -316,3 +316,10 @@
 
 @inline scalbn(x::Float64, y::Int32) = @wrap __nv_scalbn(x::double, y::i32)::double
 @inline scalbn(x::Float32, y::Int32) = @wrap __nv_scalbnf(x::float, y::i32)::float
+
+# extra
+@inline function abs(z::Complex{T}) where T
+    im = Base.imag(z)
+    re = Base.real(z)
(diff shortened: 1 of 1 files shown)


vchuravy:
I don't think we want to reimplement the entire Complex interface in CUDAnative. In the long-run we will pursue the Cassette based approach in #334 (Julia 1.3). In the short-term you can use `GPUifyLoops.contextualize`

Roger-luo:
I see. Is there any doc for this? Or if I use `@loop` macro it will add the context automatically?

vchuravy:
https://github.com/vchuravy/GPUifyLoops.jl/blob/master/examples/contextualize.jl

Roger-luo:
@vchuravy I didn't get it, there's no call to this function `contextualize` in the example. Do you mean I could also just use `@launch CUDA()`? and it will overdub a context to my source code?

vchuravy:
Ah right, sorry about that. Yes you could just use `@launch CUDA()` or see here for a example https://github.com/vchuravy/GPUifyLoops.jl/blob/760700e12f88c4bebb072d5f0453bed1d24692a3/test/test.jl#L53-L57

Roger-luo:
Ah I see thanks.

Issue #19, opened 2016-11-09
User-friendly at-cuprintf


maleadt:
`@cuprintf` is nice, but has horrible internals. Constant strings should be emitted as module-scope globals, not heap-allocated objects...

PR #439, opened 2019-07-21
Add a wrapper for at-cuprintf.

maleadt:
Just a temporary improvement, we should really work towards getting `show(::IO, x)` to work.
Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/19


maleadt:
CI failure is Documenter.jl related.

PR #441, opened 2019-07-26
Escape $ in docstrings

mortenpi:
You can do interpolation into docstrings, so it would currently show up as `@cuprint "Hello, World 42\n"` in the manual, which would, I'd argue, defeats the purpose of the example :slightly_smiling_face:

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,3 +1,7 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+
+[compat]
+Documenter = "~0.22.6"
+
--- a/src/device/cuda/output.jl
+++ b/src/device/cuda/output.jl
@@ -19,9 +19,9 @@ export @cuprintf
 end
 
 """
-Print a formatted string in device context on the host standard output:
-
     @cuprintf("%Fmt", args...)
+
+Print a formatted string in device context on the host standard output.
 
 Note that this is not a fully C-compliant `printf` implementation; see the CUDA
 documentation for supported options and inputs.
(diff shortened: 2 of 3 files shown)


mortenpi:
I would also actually wrap it into a ```` ```julia ```` code block to get syntax highlighting.

maleadt:
Good suggestion! So this doesn't need a new version of Documenter? Let's try this.
bors try

maleadt:
Oh I guess this still needs https://github.com/JuliaDocs/Documenter.jl/pull/1075

mortenpi:
> So this doesn't need a new version of Documenter?

Unless other docstrings actually have some problems too.. which they appear to have. JuliaDocs/Documenter.jl#1075 will stop Documenter from crashing in this situation.

maleadt:
OK, I'll have a look at the other docstrings.

maleadt:
If I add some output before the problematic call to `convert`, I see:

``````
x = ```
(::HostKernel)(args...; kwargs...)
(::DeviceKernel)(args...; kwargs...)
```

Low-level interface to call a compiled kernel, passing GPU-compatible arguments in `args`. For a higher-level interface, use [`@cuda`](@ref).

The following keyword arguments are supported:

  * `threads` (defaults to 1)
  * `blocks` (defaults to 1)
  * `shmem` (defaults to 0)
  * `config`: callback function to dynamically compute the launch configuration. should accept a `HostKernel` and return a name tuple with any of the above as fields. this functionality is intended to be used in combination with the CUDA occupancy API.
  * `stream` (defaults to the default stream)

ERROR: LoadError: MethodError: no method matching _convert_block(::Markdown.MD)
``````

This comes from:

```
"""
    (::HostKernel)(args...; kwargs...)
    (::DeviceKernel)(args...; kwargs...)

Low-level interface to call a compiled kernel, passing GPU-compatible arguments in `args`.
For a higher-level interface, use [`@cuda`](@ref).

The following keyword arguments are supported:
- `threads` (defaults to 1)
- `blocks` (defaults to 1)
- `shmem` (defaults to 0)
- `config`: callback function to dynamically compute the launch configuration.
  should accept a `HostKernel` and return a name tuple with any of the above as fields.
  this functionality is intended to be used in combination with the CUDA occupancy API.
- `stream` (defaults to the default stream)
"""
```

What's wrong with this docstring?

Issue #430, opened 2019-07-05
Can't broadcast complex Array with exp


navidcy:
(I'm not sure this issue belongs here -- please advise me accordingly.)

The issue is demonstrated below:

```julia
julia> a = rand(2, 2) + im*rand(2, 2)
2×2 Array{Complex{Float64},2}:
 0.465535+0.449849im  0.381765+0.0849708im
 0.387269+0.847888im  0.380261+0.375675im

julia> b = CuArray(a)
2×2 CuArray{Complex{Float64},2}:
 0.465535+0.449849im  0.381765+0.0849708im
 0.387269+0.847888im  0.380261+0.375675im

julia> exp.(a)
2×2 Array{Complex{Float64},2}:
   1.4344+0.692625im  1.45958+0.124321im
 0.974459+1.10455im   1.36066+0.536653im

julia> exp.(b)
ERROR: MethodError: no method matching exp(::Complex{Float64})
You may have intended to import Base.exp
Closest candidates are:
  exp(::Float32) at /home/navid/.julia/packages/CUDAnative/gJDZI/src/device/cuda/math.jl:90
  exp(::Float64) at /home/navid/.julia/packages/CUDAnative/gJDZI/src/device/cuda/math.jl:89
Stacktrace:
 [1] _broadcast_getindex at ./broadcast.jl:578 [inlined]
 [2] getindex at ./broadcast.jl:511 [inlined]
 [3] copy at ./broadcast.jl:787 [inlined]
 [4] materialize(::Base.Broadcast.Broadcasted{Base.Broadcast.ArrayStyle{CuArray},Nothing,typeof(CUDAnative.exp),Tuple{CuArray{Complex{Float64},2}}}) at ./broadcast.jl:753
 [5] top-level scope at none:0

julia>
```

ali-ramadhan:
I think in general you have to use or broadcast over `CUDAnative.exp` but then the issue is that `CUDAnative.exp` is not defined for `Complex{Float64}`...

Not sure if CUDA provides a complex exponential (or complex math), would definitely be nice to have if possible.

I have some complex exponentials that I currently calculate on the CPU then move them them to a CuArray but it's part of model setup so does not need to be fast. But if I had to calculate them frequently, then I'm not sure what I would do. I guess split the calculation up using `real()` and `imag()` but that wouldn't be ideal...

[...]

PR #445, opened 2019-08-19
Add some more complex operations

benchislett:
#430 

This may not be the way we want to do this, and I'm open to alternatives, but as I see it this is an effective solution.

If we want a more full set of operations to match those for real values, I can add those as well--just not sure how far we want to go.

### DIFF
--- a/src/device/cuda/math.jl
+++ b/src/device/cuda/math.jl
@@ -103,6 +103,9 @@
 @inline ldexp(x::Float64, y::Int32) = @wrap __nv_ldexp(x::double, y::i32)::double
 @inline ldexp(x::Float32, y::Int32) = @wrap __nv_ldexpf(x::float, y::i32)::float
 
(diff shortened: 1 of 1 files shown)


PhilipVinc:
Is there any reason why this is not being considered for merging? 
I would find It very useful.

On a related note, it might be useful to also add `log1p` to the list.

maleadt:
> Is there any reason why this is not being considered for merging?
> I would find It very useful.

Sorry, I forgot about this PR. Could you add some tests?

PhilipVinc:
@benchislett if you don't have time to do it I can take care of the tests.
I'd love this to be merged relatively soon.

PhilipVinc:
@benchislett as you did not answer I took the liberty of taking your commit and adding tests
PR #445.

If you prefer to handle this yourself, just tell me and I can close the PR.

PR #448, opened 2019-08-24
Fix argument types of isnan and isinf.

mwarusz:


maleadt:
Great, thanks!
bors try

Issue #420, opened 2019-06-12
Fix shfl


maleadt:
Needs tests for `_sync`, and probably use LLVM intrinsics. Also seems to be broken:

```
Lucas Wilcox [8:38 PM]
I can’t seem to get `shfl_down_sync` to work.  I tried replacing `shfl_down(d[t], n÷2)` with `shfl_down_sync(d[t], n÷2, 32, 0xffffffff)` in this test https://github.com/JuliaGPU/CUDAnative.jl/blob/be62124bbc4d40760e868c0ce51ac20e42958a34/test/device/cuda.jl#L424 and am seeing compiler errors.  Does anybody have an idea of what I am doing wrong or how to fix things?
Lucas Wilcox [8:51 PM]
The error message I get is
```

```julia> include("shfl_verify.jl")
ERROR: LoadError: InvalidIRError: compiling kernel(CuDeviceArray{Int32,1,CUDAnative.AS.Global}, Int64) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to split_value_invocation(op::Function, val, args...) in CUDAnative at /home/lucas/.julia/packages/CUDAnative/q7ezN/src/device/tools.jl:187)
Stacktrace:
 [1] macro expansion at /home/lucas/.julia/packages/CUDAnative/q7ezN/src/device/tools.jl:218
 [2] recurse_value_invocation at /home/lucas/.julia/packages/CUDAnative/q7ezN/src/device/tools.jl:217
 [3] shfl_down_sync at /home/lucas/.julia/packages/CUDAnative/q7ezN/src/device/cuda/warp_shuffle.jl:75
 [4] kernel at /home/lucas/research/code/Heptapus.jl/examples/reduce/shfl_verify.jl:9
```

PR #464, opened 2019-09-25
Improvements to shfl

maleadt:
Use LLVM intrinsics, and have `shfl_idx` use 1-based indexing.


maleadt:
cc @lcw this should fix your issue in #420

PR #466, opened 2019-09-28
Add some more complex operations - Take 2

PhilipVinc:
This is the commit from @benchislett in PR #445, with the addition of the `angle` and `log` operations and several tests for all functions included in the PR.

@benchislett If you prefer to take my commits into your PR, please feel free to do it. I'm doing this just because I have some urgency in having those merged.

On a side note:
I would love to add more functions (notably, `log1p` and `expm1`, `sort`) but I am not sure how to do this. Giving a look at [thrust](https://github.com/thrust/thrust/blob/7df7efe3542a0ab549530bc478467320467e0094/thrust/detail/complex/csqrt.h) they have a bunch of if/else logic like in base Julia. Is this even a good thing to do on the GPU?

cc @maleadt


maleadt:
Thanks!
bors try

FWIW, it's appreciated if you rebase instead of merge, to avoid many merge commits in the history.

> I would love to add more functions (notably, `log1p` and `expm1`, `sort`) but I am not sure how to do this. Giving a look at [thrust](https://github.com/thrust/thrust/blob/7df7efe3542a0ab549530bc478467320467e0094/thrust/detail/complex/csqrt.h) they have a bunch of if/else logic like in base Julia. Is this even a good thing to do on the GPU?

If there's a branchless equivalent, that's preferred. Small branches don't hurt though, because that will just compile down to predicated instructions. But it's better to avoid large branches that would result in lots of thread divergence, yes.

PhilipVinc:
I'll keep that in mind for next time, sorry!

Do you know anywhere I could look for ideas on GPU implementations of those functions?

PhilipVinc:
Do you need anything else or this is good for you?

maleadt:
Nope, looks good, thanks!
bors r+

> Do you know anywhere I could look for ideas on GPU implementations of those functions?

For `sort`, you can use a sorting network (e.g. bitonic sort). I don't have specific suggestions for the other calls you mention (expm1, log1p).

PR #467, opened 2019-09-30
Making @atomic robust

### LINKED ISSUE #453: @atomic does not error nicely
```
julia> @cuda ((a)->(CUDAnative.@atomic a[2] = 42; nothing))(CuArray{Int}(undef, 1))
ERROR: LoadError: type Int64 has no field head
```

UmashankarTriforce:
This PR aims to handle various errors encountered by `@atomic` and report them nicely
Fixes #453

### DIFF
--- a/src/device/cuda/atomics.jl
+++ b/src/device/cuda/atomics.jl
@@ -319,7 +319,7 @@ macro atomic(ex)
     if ex.head == :(=)
         ref = ex.args[1]
         rhs = ex.args[2]
-        if typeof(rhs) <: Number || typeof(rhs) <: String
+        if typeof(rhs) <: Number || typeof(rhs) <: String || typeof(rhs) <: Symbol
             error("right-hand side of an @atomic assignment cannot be a literal")
         end
         rhs.head == :call || error("right-hand side of an @atomic assignment should be a call")
--- a/src/device/cuda/atomics.jl
+++ b/src/device/cuda/atomics.jl
@@ -299,6 +299,10 @@ const inplace_ops = Dict(
     :(⊻=) => :(⊻)
 )
 
+struct AtomicError <: Exception
+    msg::AbstractString
+end
+
 """
     @atomic a[I] = op(a[I], val)
     @atomic a[I] ...= val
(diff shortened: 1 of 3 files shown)


UmashankarTriforce:
I would like to know what other conditions I need to check for. Hence I have added `[WIP]` in PR.

UmashankarTriforce:
Where and how do I add a test?

maleadt:
Have a look at the `tests/` folder, there's a bunch of existing ones for `@atomic`.

UmashankarTriforce:
I have added a section for `exception` test and accomodated the previous failing case. Let me know if I have missed out on something

src/device/cuda/atomics.jl
```diff
@@ -319,6 +319,9 @@ macro atomic(ex)
     if ex.head == :(=)
         ref = ex.args[1]
         rhs = ex.args[2]
+        if typeof(rhs) <: Number || typeof(rhs) <: String || typeof(rhs) <: Symbol
```
maleadt: That's not what I meant, instead of checking against a blacklist of non-Expr types, it's easier to (whitelist) check for what's allowed: `Expr`.

test/device/cuda.jl
```diff
@@ -981,6 +981,18 @@ end
...
+        a = CuArray([typemax(T)])
+        function kernal(T, a)
+            @atomic a[1] = 1
+            @atomic a[1] = "1"
+            @atomic a[1] = i
+            return
+        end
+        @cuda threads=32 kernal(T, a)
+        @test Array(a)[1] == 1
+    end
```
maleadt: Did you run these tests? They are intended to fail, so you can't just run them (the test would fail). Instead, you need to `test_throws`. Also, you can't put multiple failing atomic expressions in a single test because it'll bail out at the first error.
I'll push a fix, since it's somewhat finicky to use `test_throws` with macros.

maleadt:
bors try

src/device/cuda/atomics.jl
UmashankarTriforce: Ahh okay I'll fix this

test/device/cuda.jl
UmashankarTriforce: I did test them, but they didn't throw any error

maleadt:
Let's merge this. Thanks for the PR!

PR #473, opened 2019-10-16
Add autogenerated wrappers for CUPTI and NVTX

maleadt:


maleadt:
cc @vchuravy folding NVTX.jl into here, I don't think it makes much sense to have a separate package for that.

maleadt:
CI will suffer from this change until we reboot cyclops.

PR #474, opened 2019-10-16
Use a module exception flag instead of trapping the GPU.

maleadt:
This is a promising change, making it possible to throw exceptions without killing the GPU:

```julia
julia> @cuda ((a)->(a[2]=1;return))(CuArray([1]))

julia> ERROR: a exception was thrown during kernel execution.
       Run Julia on debug level 2 for device stack traces.
julia> 

julia> synchronize()
ERROR: KernelException: exception thrown during kernel execution on device Quadro RTX 5000
Stacktrace:
 [1] check_exceptions() at /home/tim/Julia/pkg/CUDAnative/src/exceptions.jl:50
 [2] check_exception_hook(::Symbol) at /home/tim/Julia/pkg/CUDAnative/src/exceptions.jl:60
 [3] macro expansion at /home/tim/Julia/pkg/CUDAdrv/src/error.jl:94 [inlined]
 [4] cuCtxSynchronize() at /home/tim/Julia/pkg/CUDAdrv/src/libcuda.jl:156
 [5] synchronize() at /home/tim/Julia/pkg/CUDAdrv/src/context.jl:185
 [6] top-level scope at REPL[8]:1

julia> @cuda ((a)->(a[2]=1;return))(CuArray([1,2]))

julia> synchronize()
```

Illegal memory errors etc. still break CUDA, but many of those should be guarded by bounds checks and our IR verifier.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -21,8 +21,10 @@ uuid = "3895d2a7-ec45-59b8-82bb-cfc6a382f9b3"
 version = "1.2.0"
 
 [[CUDAdrv]]
-deps = ["CUDAapi", "Libdl", "Printf"]
-git-tree-sha1 = "9ce99b5732c70e06ed97c042187baed876fb1698"
+deps = ["CEnum", "CUDAapi", "Libdl", "Printf"]
+git-tree-sha1 = "5d739aece437581b5d44783db90c2488973f8721"
+repo-rev = "master"
+repo-url = "https://github.com/JuliaGPU/CUDAdrv.jl.git"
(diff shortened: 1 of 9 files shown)


maleadt:
I tried this a while ago already, https://github.com/JuliaGPU/CUDAnative.jl/commits/tb/exceptions, but we didn't have pinned memory wrapped back then which made it hard to preserve asynchronous execution.

vchuravy:
<3 . Since we generate one flag per module and we have one module per kernel function extracting the name seems not to complicated.

maleadt:
This seems to work surprisingly well. Let's just hit the big green button and see how it fares.

PR #502, opened 2019-11-26
Don't always time operations.

maleadt:


maleadt:
Why is this not running CI...
bors try

PR #505, opened 2019-11-27
use map instead of broadcast

SimonDanisch:
This gets rid of some allocations and make the result inferrable


maleadt:
Thanks! By how much did this reduce allocations?
I guess the `Core.Typeof` broadcast below does get properly inferred at compile time?

PR #506, opened 2019-11-27
Add addrspacecast function

thomasfaingnaert:
This PR adds an `addrspacecast` function that wraps LLVM's `addrspacecast`. This is useful to subtract the base address of e.g. the shared AS window from the value stored in `DevicePtr`s.

```julia
julia> code_llvm(addrspacecast, Tuple{CUDAnative.DevicePtr{Float32, AS.Generic}})

;  @ /home/tfaingna/.julia/dev/CUDAnative/src/device/pointer.jl:284 within `addrspacecast'
define i64 @julia_addrspacecast_18076(i64) {
top:
; ┌ @ /home/tfaingna/.julia/dev/CUDAnative/src/device/pointer.jl:284 within `macro expansion'
   ret i64 %0
; └
}

julia> code_llvm(addrspacecast, Tuple{CUDAnative.DevicePtr{Float32, AS.Shared}})

;  @ /home/tfaingna/.julia/dev/CUDAnative/src/device/pointer.jl:284 within `addrspacecast'
define i64 @julia_addrspacecast_17283(i64) {
top:
; ┌ @ /home/tfaingna/.julia/dev/CUDAnative/src/device/pointer.jl:284 within `macro expansion' @ /home/tfaingna/.julia/packages/LLVM/ICZSf/src/interop/base.jl:52
   %1 = inttoptr i64 %0 to float*
   %2 = addrspacecast float* %1 to float addrspace(3)*
   %3 = ptrtoint float addrspace(3)* %2 to i64
; │ @ /home/tfaingna/.julia/dev/CUDAnative/src/device/pointer.jl:284 within `macro expansion'
   ret i64 %3
; └
}
```

This is in reference to point 2 of https://github.com/JuliaGPU/CUDAnative.jl/pull/494#issuecomment-554441290, but since the change is rather self-contained, I decided to open a new PR for it to separate it from WMMA.
This might also prove useful to transition some `@generated` functions over to `ccall`, say https://github.com/JuliaGPU/CUDAnative.jl/blob/b55dee83f0cf4e51c937e449cb30f5e0101c846e/src/device/pointer.jl#L219.

### DIFF
--- a/src/device/pointer.jl
+++ b/src/device/pointer.jl
@@ -276,3 +276,32 @@ end
 @inline unsafe_cached_load(p::DevicePtr{T,AS.Global}, i::Integer=1, args...) where {T} =
     recurse_pointer_invocation(unsafe_cached_load, p+sizeof(T)*Int(i-one(i)),
                                CachedLoadPointers, 1, args...)
+
+export addrspacecast
+
+@generated function addrspacecast(p::DevicePtr{T, AS}) where {T, AS}
+    # types
+    eltyp = convert(LLVMType, T)
+    T_ptr = convert(LLVMType, DevicePtr{T, AS})
+
+    T_actual_ptr = LLVM.PointerType(eltyp)
+    T_actual_ptr_as = LLVM.PointerType(eltyp, convert(Int, AS))
+
+    # create function
+    param_types = [T_ptr]
+    llvm_f, _ = create_function(T_ptr, param_types)
+
+    # generate LLVM IR
+    Builder(JuliaContext()) do builder
+        entry = BasicBlock(llvm_f, "entry", JuliaContext())
+        position!(builder, entry)
+
+        ptr = inttoptr!(builder, parameters(llvm_f)[1], T_actual_ptr)
+        ptr_with_as = addrspacecast!(builder, ptr, T_actual_ptr_as)
+        ret = ptrtoint!(builder, ptr_with_as, LLVM.Int64Type(JuliaContext()))
+
+        ret!(builder, ret)
+    end
+
+    call_function(llvm_f, DevicePtr{T, AS}, Tuple{DevicePtr{T, AS}}, :((p,)))
+end


maleadt:
Right, but emitting an `addrspacecast` in LLVM -> `cvta` in PTX that subtracts the offset from a generic pointer (do you have source for it actually doing that, and modifying the pointer value? I thought that the casts were merely providing state space information to the assembler to know whether it needs to perform a pointer windowing check at run time) is only one part. In LLVM, actually having the addrspace information at the time of a memory operation allows it to emit instructions that are tagged with the state space. That information gets lost with your approach:

```
julia> foo(ptr::DevicePtr{T}) where T = unsafe_load(reinterpret(Ptr{T}, addrspacecast(ptr)))

julia> code_llvm(foo, (DevicePtr{Int8,AS.Shared},))

;  @ REPL[16]:1 within `foo'
define i8 @julia_foo_17434(i64) {
top:
; ┌ @ REPL[2]:5 within `addrspacecast'
; │┌ @ REPL[2]:5 within `macro expansion' @ /home/tim/Julia/pkg/LLVM/src/interop/base.jl:52
    %1 = inttoptr i64 %0 to i8*
    %2 = addrspacecast i8* %1 to i8 addrspace(3)*
    %3 = ptrtoint i8 addrspace(3)* %2 to i64
; └└
; ┌ @ pointer.jl:105 within `unsafe_load' @ pointer.jl:105
   %4 = inttoptr i64 %3 to i8*
   %5 = load i8, i8* %4, align 1
; └
  ret i8 %5
}

julia> CUDAnative.code_ptx(foo, (DevicePtr{Int8,AS.Shared},))
//
// Generated by LLVM NVPTX Back-End
//

.version 6.0
.target sm_70
.address_size 64

        // .globl       julia_foo_4     // -- Begin function julia_foo_4
                                        // @julia_foo_4
.visible .func  (.param .b32 func_retval0) julia_foo_4(
        .param .b64 julia_foo_4_param_0
)
{
        .reg .b32       %r<2>;
        .reg .b64       %rd<3>;

// %bb.0:                               // %top
        ld.param.u64    %rd1, [julia_foo_4_param_0];
        cvta.to.shared.u64      %rd2, %rd1;
        ld.u8   %r1, [%rd2];
        st.param.b32    [func_retval0+0], %r1;
        ret;
}
                                        // -- End function
```

Wheras the current deviceptr-specific pointer load performs the cast right before the load:

```
julia> bar(ptr::DevicePtr{T}) where T = unsafe_load(ptr)

julia> code_llvm(bar, (DevicePtr{Int8,AS.Shared},))

;  @ REPL[19]:1 within `bar'
define i8 @julia_bar_17436(i64) {
top:
; ┌ @ /home/tim/Julia/pkg/CUDAnative/src/device/pointer.jl:132 within `unsafe_load' @ /home/tim/Julia/pkg/CUDAnative/src/device/pointer.jl:132 @ /home/tim/Julia/pkg/CUDAnative/src/device/pointer.jl:132
; │┌ @ /home/tim/Julia/pkg/CUDAnative/src/device/pointer.jl:132 within `macro expansion' @ /home/tim/Julia/pkg/LLVM/src/interop/base.jl:52
    %1 = inttoptr i64 %0 to i8*
    %2 = addrspacecast i8* %1 to i8 addrspace(3)*
    %3 = load i8, i8 addrspace(3)* %2, align 1
; └└
  ret i8 %3
}

julia> CUDAnative.code_ptx(bar, (DevicePtr{Int8,AS.Shared},))
//
// Generated by LLVM NVPTX Back-End
//

.version 6.0
.target sm_70
.address_size 64

        // .globl       julia_bar_5     // -- Begin function julia_bar_5
                                        // @julia_bar_5
.visible .func  (.param .b32 func_retval0) julia_bar_5(
        .param .b64 julia_bar_5_param_0
)
{
        .reg .b32       %r<2>;
        .reg .b64       %rd<3>;

// %bb.0:                               // %top
        ld.param.u64    %rd1, [julia_bar_5_param_0];
        cvta.to.shared.u64      %rd2, %rd1;
        ld.shared.u8    %r1, [%rd2];
        st.param.b32    [func_retval0+0], %r1;
        ret;
}
                                        // -- End function
```

And you can't expect LLVM to optimize away the `%3 = ptrtoint i8 addrspace(3)* %2 to i64` -> `%4 = inttoptr i64 %3 to i8*` because that would actually change the type from `i8*` into `i8 addrspace(3)*`. Although that's what the AS interference pass does...

thomasfaingnaert:
This PR is mainly to replace the `generic_to_shared` function here: https://github.com/JuliaGPU/CUDAnative.jl/pull/494/files#diff-48152fae1cd73f1903d63337b3e02824R52.

Here's an example where `addrspacecast` subtracts the window base:
```julia
           mem = @cuStaticSharedMem(Float32, 16)
           ptr = mem.ptr
           @cuprintf("%ld\n", convert(Int64, ptr))
           
           ptr_asc = addrspacecast(ptr)
           @cuprintf("%ld\n", convert(Int64, ptr_asc))
           
           return
       end

@cuda kernel()
```
which prints:
```
140066869477376
0
```

maleadt:
Good to know the pointer value actually changes!

On another note, it would be interesting if Julia itself would know about the pointer address space. Then we could just adapt the pointer value as you do here, inform Julia about the address space, and continue our lives. But changing `Ptr{T}` to `Ptr{T,AS}` proved much too invasive (and pretty hard post-1.0). I guess it would be possible to somehow hack it in, as long as the compiler knows about it and preserves the type when `inttoptr`ing back to a pointer for use by LLVM. That would make it possible to reuse the pointer intrinsics, but wouldn't buy us seamless reuse because base Julia performs load with an alignment of 1 (which is horribly slow on GPUs). This is an actual problem with e.g. StaticArrays.

Issue #408, opened 2019-05-20
Cannot assign variable to module... ?


ChrisRackauckas:
```julia
using Flux, CuArrays
x = cu(rand(4))
p = param(cu(rand(4)))
_f = function (p)
  convert.(eltype(p),x)
end
_f(p)
```

@vchuravy That's a fun one.

maleadt:
https://github.com/JuliaLang/julia/issues/33347

ChrisRackauckas:
I don't see how that's related? `convert.(Float32,x)` seems like it should work on a CuArray?

```julia
using CuArrays
x = cu(rand(4))
convert.(Float32,x)
```

is a better MWE.

maleadt:
> I don't see how that's related?

It literally throws exactly that error when executing your MWE, so how can it _not_ be related?

The reason it triggers that code path in the first place is because you're passing `Float32` to Base.broadcast, which encapsulates it in a `Ref` and dereferences (i.e. uses) that ref on the device. Using non-bitstypes isn't allowed. The following would work:

```julia
julia> map(x->convert(Float32,x), x)
4-element CuArray{Float32,1,Nothing}:
 0.9483574 
 0.7480514 
 0.8318354 
 0.23302273
```

ChrisRackauckas:
Oh, I still don't see why that would call fieldtypes on a module, but I'll take your word for it.

How come Float32 wouldn't be a bitstype though? It's a singleton which I would've assumed would've worked.

maleadt:
```
julia> isbits(Float32)
false
```

Sure, it's a singleton, but we don't special case that currently. It's allowed if the code gets specialized on it and the value itself isn't used, but the `Ref` breaks that.

> Oh, I still don't see why that would call fieldtypes on a module, but I'll take your word for it.

Some debugging that introspects types to see why they aren't bits.

PR #521, opened 2019-12-23
Limit deptch of nonisbits-explanation tree and work around Julia issue.

maleadt:
Fixes (the original issue from) #408, works around JuliaLang/julia#33347.


KristofferC:
Ref https://github.com/JuliaLang/julia/issues/33347.

PR #522, opened 2019-12-23
Fixes for Julia 1.4: disable DI on LLVM 8.0.

maleadt:


maleadt:
With LLVMDebugEmissionKindFullDebug CUDA segfaults, using LLVMDebugEmissionKindLineTablesOnly LLVM asserts, using the following script on `-g2`:

```
using CUDAnative, CUDAdrv

@noinline foo(a, i) = a[1] = i
bar(a) = (foo(a, 42); nothing)

ptr = CUDAnative.DevicePtr{Int,AS.Global}(0)
arr = CuDeviceArray{Int,1,AS.Global}((0,), ptr)

@cuda bar(arr)
CUDAdrv.synchronize()

```

PR #532, opened 2020-01-24
Implementation of atomic floating-point operations using atomic_cas!

yuehhua:


maleadt:
Awesome, thanks! This is really useful to generalize our atomic operations.

Some comments:
- I think this is only on sm_6x, so we should guard the tests
- you are using the low-level LLVM intrinsics (`llvm.nvvm.atomic.cas.gen`) while we normally use NVVM intrinsics (`llvm.nvvm.atomic.load.add`) that get lowered to those system-wide atomics. I didn't find one for CAS though, did you?
- even better though would be to rely on LLVM to generate even the NVVM intrinsics. Did you try generating an `CmpXchg` intstruction? So using `LLVM.atomic_cmpxhg!` (just like we use `atomic_rmw!` for some others)?

yuehhua:
> * I think this is only on sm_6x, so we should guard the tests

So.............how exactly we have to do?

> * you are using the low-level LLVM intrinsics (`llvm.nvvm.atomic.cas.gen`) while we normally use NVVM intrinsics (`llvm.nvvm.atomic.load.add`) that get lowered to those system-wide atomics. I didn't find one for CAS though, did you?

I found atomic cas [here](https://github.com/microsoft/llvm-1/blob/44486be0a1dfb6e34fdc668998e3b051a57b48d1/test/CodeGen/NVPTX/atomics-with-scope.ll#L97). I checked there are two kind s of commands, one for thread block level and another for system level. I just use the system level one.
And I found they only support `Float32` and `Float64`. They works!

> * even better though would be to rely on LLVM to generate even the NVVM intrinsics. Did you try generating an `CmpXchg` intstruction? So using `LLVM.atomic_cmpxhg!` (just like we use `atomic_rmw!` for some others)?

I tried the generating code from LLVM, just the same with other atomic operations relying on `atomic_rmw!`. I used `LLVM.atomic_cmpxhg!` and set several ordering stuffs, but it doesn't work. I got unsupported dynamic kernel advocating. After that I turn to use NVVM intrinsics directly.

maleadt:
Looks like I can't push to your fork; I created a new PR here: https://github.com/JuliaGPU/CUDAnative.jl/pull/534

PR #534, opened 2020-01-28
Implement atomic_cas!

maleadt:
Amend https://github.com/JuliaGPU/CUDAnative.jl/pull/532 (which didn't have "push from maintainers" enabled).


maleadt:
@yuehhua please verify that my changes are compatible with your use -- I assume so because everything still passes tests. This should be good to go then.

yuehhua:
@maleadt I reviewed. It works and tests passed, so it's good.

maleadt:
Great, thanks for the contribution!

PR #537, opened 2020-01-30
Fix compatibility with older GPUArrays.

maleadt:
Old versions of GPUArrays invoke `_shmem` with an integer id; make sure those are unique to pass `ptxas`.

Fixes issue introduced by https://github.com/JuliaGPU/CUDAnative.jl/pull/533


haampie:
This solves my issue! :)

Issue #553, opened 2020-02-10
CUDAnative.fma performance problems


jkozdon:
On the TitanV and V100 I am seeing a degredation of performance when using `CUDAnative.fma`. The following is modification of the [peakflops] example, and shows that when `CUDAnative.fma` is used the code performance is significantly worse than no fma.

```julia
using CUDAdrv, CUDAnative, CuArrays

using Test

"Dummy kernel doing 100 FMAs."
function kernel_100fma(a, b, c, out)
    i = (blockIdx().x-1) * blockDim().x + threadIdx().x
    @inbounds a_val = a[i]
    @inbounds b_val = b[i]
    @inbounds c_val = c[i]

    for j in 1:33
        a_val = CUDAnative.fma(a_val, b_val, c_val)
        b_val = CUDAnative.fma(a_val, b_val, c_val)
        c_val = CUDAnative.fma(a_val, b_val, c_val)
    end

    @inbounds out[i] = CUDAnative.fma(a_val, b_val, c_val)

    return
end

function kernel_100(a, b, c, out)
    i = (blockIdx().x-1) * blockDim().x + threadIdx().x
    @inbounds a_val = a[i]
    @inbounds b_val = b[i]
    @inbounds c_val = c[i]

    for j in 1:33
        a_val = a_val *  b_val + c_val
        b_val = a_val *  b_val + c_val
        c_val = a_val *  b_val + c_val
    end

    @inbounds out[i] = a_val * b_val + c_val

    return
end

function peakflops(n::Integer=5000, dev::CuDevice=CuDevice(0))
    device!(dev) do
        dims = (n, n)
        a = round.(rand(Float32, dims) * 100)
        b = round.(rand(Float32, dims) * 100)
        c = round.(rand(Float32, dims) * 100)
        out = similar(a)

        d_a = CuArray(a)
        d_b = CuArray(b)
        d_c = CuArray(c)
        d_out = CuArray(out)

        len = prod(dims)
        threads = min(len, 1024)
        blocks = len ÷ threads

        # warm-up
        @cuda kernel_100fma(d_a, d_b, d_c, d_out)
        synchronize()

        secs = CUDAdrv.@elapsed begin
            @cuda blocks=blocks threads=threads kernel_100fma(d_a, d_b, d_c, d_out)
        end
        flopcount = 200*len
        fma_flops = flopcount / secs

        @cuda kernel_100(d_a, d_b, d_c, d_out)
        synchronize()

[...]

PR #554, opened 2020-02-11
Run the inliner at the end of the pipeline.

maleadt:
This works around its inability to deal with Julia's operand bundles.
Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/553

### DIFF
--- a/src/compiler/optim.jl
+++ b/src/compiler/optim.jl
@@ -15,55 +15,36 @@ function optimize!(job::CompilerJob, mod::LLVM.Module, entry::LLVM.Function)
     #
     # NOTE: we need to use multiple distinct pass managers to force pass ordering;
     #       intrinsics should never get lowered before Julia has optimized them.
-    if VERSION < v"1.2.0-DEV.375"
-        # with older versions of Julia, intrinsics are lowered unconditionally so we need to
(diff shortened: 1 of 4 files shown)


src/bindeps.jl
```diff
@@ -0,0 +1,222 @@
+# discovering binary CUDA dependencies
```
vchuravy: Accidental checkin?

src/compiler/optim.jl
```diff
@@ -15,56 +15,39 @@ function optimize!(job::CompilerJob, mod::LLVM.Module, entry::LLVM.Function)
...
-                late_lower_gc_frame!(pm)
-            end
+        # lower intrinsics
+        add!(pm, FunctionPass("LowerGCFrame", lower_gc_frame!))
+        aggressive_dce!(pm) # remove dead uses of ptls
+        add!(pm, ModulePass("LowerPTLS", lower_ptls!))
 
-            run!(pm, mod)
+        # the Julia GC lowering pass also has some clean-up that is required
+        if VERSION >= v"1.2.0-DEV.531"
```
vchuravy: Compat is at `julia = "1.3"`

PR #579, opened 2020-02-25
Rework pointer/array interfaces, simplify shuffle implementation

maleadt:
Introduce pointerref/pointerset, arrayset/arrayref, simplify methods building on that (like unsafe_load).

cc @vchuravy, this removes some fragile capabilities of doing ldg/shuffle on arbitrary values and types that aren't supported by the hardware. I figure it's better to keep the compiler/stdlib simple and expect the user to have some smarts about which operations work on the GPU and which do not.


vchuravy:
FIne by me :)

maleadt:
@vchuravy I similarly moved from `ldg` to `Const` here, looks OK?

src/device/array.jl
```diff
@@ -66,44 +66,68 @@ Base.length(g::CuDeviceArray) = prod(g.shape)
...
 """
-@inline function ldg(A::CuDeviceArray{T}, index::Integer) where {T}
-    # FIXME: this only works on sm_35+, but we can't verify that for now
-    @boundscheck checkbounds(A, index)
-    align = Base.datatype_alignment(T)
-    unsafe_cached_load(pointer(A), index, Val(align))::T
+struct Const{T,N,AS} <: DenseArray{T,N}
+    a::CuDeviceArray{T,N,AS}
 end
+Base.Experimental.Const(A::CuDeviceArray) = Const(A)
```
vchuravy: I would leave this out since otherwise one could expect that it works with `Base.arrayref_const`

maleadt: Why's that? It just treats `Const` as the interface, implementing the Base method but still returning a CUDAnative.Const, so there's no use of `Base.arrayref_const`. Since the Base documentation mentions this to be an API, I think this is the correct way to go.

PR #580, opened 2020-02-27
use --startup=no in spawned tests

MasonProtter:
The test suite involves spinning up new julia instances, but these can end up loading the users startup.jl which is undesirable.


MasonProtter:
This resolves errors I was having running the test suite locally.

maleadt:
Thanks!
bors r+

Issue #459, opened 2019-09-23
265-fix is only partial


maleadt:
Not sure how I only ran into this now...

```julia
# used by CUDAnative to determine whether to recompile a kernel
julia> function method_age(f, tt)::UInt
           for m in Base._methods(f, tt, 1, typemax(UInt))
               if VERSION >= v"1.2.0-DEV.573"
                   return m[3].primary_world
               else
                   return m[3].min_world
               end
           end
           throw(MethodError(f, tt))
       end

# example kernel that uses a method
julia> foo() = bar()
julia> bar() = 1

julia> method_age(foo, Tuple{})
0x00000000000065d2

julia> method_age(bar, Tuple{})
0x00000000000065d3

# redefinition of called function
julia> bar() = 2

# its age changes...
julia> method_age(bar, Tuple{})
0x00000000000065d4

# but not the age of the caller
julia> method_age(foo, Tuple{})
0x00000000000065d2
```

@vtjnash what is the correct way to invalidate a compiled function based on whether it or any called function has changed?

maleadt:
I think I managed to come up with something that embeds proper backedges, based on https://github.com/JuliaLang/julia/pull/32237:

```julia
using Core.Compiler: retrieve_code_info, CodeInfo, MethodInstance, SSAValue, SlotNumber
using Base: _methods_by_ftype

# cached compilation and kernel execution
const compilecache = Dict{Int,Nothing}()
function cufunction(id, f, args...)
    get!(compilecache, id) do
        mi, ci = compilerequests[id]
        delete!(compilerequests, id)
        println("compile: $mi")
        nothing
    end
    f(args...)
end

# generate code to call cufunction from generator(self, f, args)
globalUnique = 0
const compilerequests = Dict()
function emit_driver_call(mi, nargs, sparams)
    ci = retrieve_code_info(mi)
    @assert isa(ci, CodeInfo)

    # generate a unique id to represent this compilation
    # NOTE: we don't embed the method and code info into the call to cufunction
    #       to prevent dynamic dispatch and significantly speed-up the fast path
    global globalUnique
    id = (globalUnique += 1)
    compilerequests[id] = (mi, ci)

    # empty the function
    new_ci = copy(ci)
    empty!(new_ci.code)
    empty!(new_ci.codelocs)
    empty!(new_ci.linetable)
    empty!(new_ci.ssaflags)

    # prepare the slots
[...]

PR #581, opened 2020-02-27
Fix method invalidation

maleadt:
Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/459 by setting `ci.edges` (from https://github.com/JuliaLang/julia/pull/32237) AND adding a fake call, both seem required to fool Julia.

This regresses launch performance, which I plan to look into soon.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -103,16 +103,11 @@ A device-side launch, aka. dynamic parallelism, is similar but more restricted:
 """
 macro cuda(ex...)
     # destructure the `@cuda` expression
-    if length(ex) > 0 && ex[1].head == :tuple
-        error("The tuple argument to @cuda has been replaced by keywords: `@cuda threads=... fun(args...)`")
-    end
     call = ex[end]
     kwargs = ex[1:end-1]
 
     # destructure the kernel call
-    if call.head != :call
-        throw(ArgumentError("second argument to @cuda should be a function call"))
-    end
+    Meta.isexpr(call, :call) || throw(ArgumentError("second argument to @cuda should be a function call"))
     f = call.args[1]
     args = call.args[2:end]
 
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -61,23 +61,6 @@ function assign_args!(code, args)
     return vars, var_exprs
 end
 
-# fast lookup of global world age
-world_age() = ccall(:jl_get_tls_world_age, UInt, ())
-
-# slow lookup of local method age
-function method_age(f, t)::UInt
-    for m in Base._methods(f, t, 1, typemax(UInt))
-        if VERSION >= v"1.2.0-DEV.573"
-            return m[3].primary_world
-        else
-            return m[3].min_world
-        end
-    end
-
-    tt = Base.to_tuple_type(t)
-    throw(MethodError(f, tt))
-end
-
 
 ## high-level @cuda interface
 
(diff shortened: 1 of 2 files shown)


maleadt:
Valentin pointed out we shouldn't need to set the edges here, and the fake call should suffice because unlike Cassette we actually call the child methods. (Conversely, shouldn't setting the edge explicitly on the outer method work too and not require a fake call?)

Anyway, something fishy is happening, because not adding edges and only doing the call breaks our tests, or more succinctly:

```
using CUDAnative, CuArrays, CUDAdrv

arr = CuArray(zeros(Int))

doit(ptr) = (unsafe_store!(ptr, 0); @cuprintln(1))

function kernel(ptr)
    doit(ptr)
    return
end

@cuda kernel(pointer(arr))

doit(ptr) = (unsafe_store!(ptr, 0); @cuprintln(2))

@cuda kernel(pointer(arr))

synchronize()
```

Maybe something's wrong with the fake call? Replacing it with the following (for the above example) makes invalidation work without the edge:

```
function fake_call(f, tt)
    opaque_false[] || return
    args = [Ref{CUDAnative.DevicePtr{Int64,CUDAnative.AS.Generic}}()[]]
    f(args...)
end
```

However, starting to generalize that breaks as soon as I do something in the generated version of this method:

```
@generated function fake_call(f, tt)
    adding_this_statement_breaks_invalidation = [:(Ref{$T}()[]) for T in tt.parameters[1].parameters]
    quote
        opaque_false[] || return
        args = [Ref{CUDAnative.DevicePtr{Int64,CUDAnative.AS.Generic}}()[]]
        f(args...)
    end
end
```

What is going on ...

maleadt:
Calling in the cavalry... @vtjnash or @Keno could you guys shed some light on this? In summary, I'm trying to get method invalidation working. This doesn't automatically work since we never call the kernel function, but take its IR and execute that. The compilation and invalidation is handled by `cufunction(t, tt)`, which in the current design is a generator that **boths** emit a code info with `ci.edges` set a la Cassette, **and** adds a fake call to `f` in its IR.
https://github.com/JuliaGPU/CUDAnative.jl/blob/b5aaeb69b2daaa906d9770f12c05f624b8b4cf2c/src/execution.jl#L318-L330
https://github.com/JuliaGPU/CUDAnative.jl/blob/b5aaeb69b2daaa906d9770f12c05f624b8b4cf2c/src/execution.jl#L453

Either one of those doesn't suffice to get proper invalidation/recompilation, but the behavior is weird and depends in strange ways on the kernel function and on the code in `fake_call` in ways I can't explain (see previous post). Shouldn't either the fake call or setting the edges get this working? Or am I doing something undefined?

src/execution.jl
```diff
@@ -334,8 +312,97 @@ end
...
+    opaque_false[] || return
+    f(Ref{Any}()[]...)
+end
+
+# actual compilation
+function cufunction_slow(f, tt, spec; name=nothing, kwargs...)
+    start = time_ns()
+
+    # generate a fake call to ensure we get recompiled upon method invalidation
+    fake_call(f)
```
vchuravy: Why does this function need to be recompiled?

src/execution.jl
```diff
@@ -334,8 +312,97 @@ end
...
 
-const agecache = Dict{UInt, UInt}()
+using Core.Compiler: retrieve_code_info, CodeInfo, MethodInstance, SSAValue, SlotNumber
+using Base: _methods_by_ftype
+
+# HACK: mechanism to generate calls that are not executed, but ensure method invalidation
+const opaque_false = Ref(false)
+function fake_call(f)
+    opaque_false[] || return
+    f(Ref{Any}()[]...)
```
vchuravy: This will give you an edge to the methodtable of `f` and only when you are lucky. I suspect that it many cases this will yield a generic call which not insert a dependency edge.

src/execution.jl
```diff
@@ -356,92 +423,51 @@ The output of this function is automatically cached, i.e. you can simply call `c
...
-
-        return compilecache[key]::HostKernel{f,tt}
-    end
+@generated function cufunction(f::Core.Function, tt::Type=Tuple{}; kwargs...)
+    # generated function that crafts a custom code info to call the actual cufunction impl.
+    # this gives us the flexibility to insert manual back edges for automatic recompilation.
+    tt = tt.parameters[1]
+
+    # get a hold of the method and code info of the kernel function
+    sig = Tuple{f, tt.parameters...}
```
vchuravy:
```suggestion
    sig = Base.signature_type(f, tt)
```

src/execution.jl
```diff
@@ -356,92 +423,51 @@ The output of this function is automatically cached, i.e. you can simply call `c
...
+    mthds = _methods_by_ftype(sig, -1, typemax(UInt))
+    length(mthds) == 1 || return (:(throw(MethodError(f,tt))))
```
vchuravy:
```suggestion
    if !Base.isdispatchtuple(tt)
        error("$tt is not a dispatch tuple")
    end
    length(mthds) == 1 || return (:(throw(MethodError(f,tt))))
```

src/execution.jl
```diff
@@ -356,92 +423,51 @@ The output of this function is automatically cached, i.e. you can simply call `c
...
+    mtypes, msp, m = mthds[1]
+    mi = ccall(:jl_specializations_get_linfo, Ref{MethodInstance}, (Any, Any, Any), m, mtypes, msp)
+    mi.def.isva && return :(error("varargs kernel methods are not supported"))
+    ci = retrieve_code_info(mi)
+    @assert isa(ci, CodeInfo)
+
+    # generate a unique id to represent this specialization
+    global specialization_counter
+    id = (specialization_counter += 1)
+    # TODO: save the mi/ci here (or embed it in the AST to pass to cufunction)
```
vchuravy: Yeah probably just pass it to cufunction along with `f, tt`

src/execution.jl
```diff
@@ -356,92 +423,51 @@ The output of this function is automatically cached, i.e. you can simply call `c
...
+    mthds = _methods_by_ftype(sig, -1, typemax(UInt))
+    length(mthds) == 1 || return (:(throw(MethodError(f,tt))))
+    mtypes, msp, m = mthds[1]
+    mi = ccall(:jl_specializations_get_linfo, Ref{MethodInstance}, (Any, Any, Any), m, mtypes, msp)
+    mi.def.isva && return :(error("varargs kernel methods are not supported"))
+    ci = retrieve_code_info(mi)
+    @assert isa(ci, CodeInfo)
+
+    # generate a unique id to represent this specialization
+    global specialization_counter
```
vchuravy: Hm, shouldn't this be the world-age of the `mi` instead of the many invocations of the generator?

maleadt: But the method age of the `mi` doesn't increase when a called method gets redefined, which was the reason for this PR in the first place.

src/execution.jl
maleadt: Ah, right. I expected the entire call chain to get recompiled, which is wrong of course. So this should be part of the generator output.

src/execution.jl
maleadt: Right, but doesn't that also mean that calling a function is inherently a fragile mechanism for adding an edge since it can always be a generic call? Explicitly adding an edge by setting it in `ci` seems like a better approach then, were it not that it does not seem to work (or I am doing something wrong).

src/execution.jl
maleadt: That results in a gcframe in the calling code though. Maybe that's a micro optimization, but it would be nice to get as clean as possible kernel launching code.

src/execution.jl
maleadt: This doesn't work here since `f` is already a typeof(function) (as we're in a generator) and signature_type does something weird then:

```
julia> Base.signature_type(sin, Tuple{})
Tuple{typeof(sin)}

julia> Base.signature_type(typeof(sin), Tuple{})
Tuple{Type{typeof(sin)}}
```

src/execution.jl
vchuravy: Yes, when I had to solve this previously, I added a call with the specific arguments. Those did not end up being generic calls since they were always duspatchtuples.

src/execution.jl
vchuravy: Hm I thought we updated the upper bound? Maybe I am simply not understanding the caching scheme you employed here.
A `@generated` function may be invoked by the compiler as often as it wants, which the condition that the same inputs yield the same output (within a worldarg range).

Instead of looking at `mi.min_world` have you looked at `mi.max_world`? That should get updated, when there is a method instance with a signature in a newer world.

maleadt: The parent world counters don't change when redefining child functions.

```
julia> function method_age(f, tt)
           sig = Base.signature_type(f, tt)
           mthds = _methods_by_ftype(sig, -1, typemax(UInt))
           Base.isdispatchtuple(tt) || return(:(error("$tt is not a dispatch tuple")))
           length(mthds) == 1 || return (:(throw(MethodError(f,tt))))
           mtypes, msp, m = mthds[1]
           mi = ccall(:jl_specializations_get_linfo, Ref{MethodInstance}, (Any, Any, Any), m, mtypes, msp)
           mi.def.isva && return :(error("varargs kernel methods are not supported"))
           ci = retrieve_code_info(mi)
           @assert isa(ci, CodeInfo)

           dump(stderr, ci)
       end
method_age (generic function with 1 method)

julia> foo() = bar()

julia> bar() = 1

julia> method_age(foo, Tuple{})
CodeInfo
  parent: MethodInstance
    def: Method
      primary_world: UInt64 0x0000000000006b26
      deleted_world: UInt64 0xffffffffffffffff
      specializations: Core.TypeMapEntry
        min_world: UInt64 0x0000000000000001
        max_world: UInt64 0xffffffffffffffff
  min_world: UInt64 0x0000000000000001
  max_world: UInt64 0xffffffffffffffff

julia> method_age(bar, Tuple{})
CodeInfo
  parent: MethodInstance
    def: Method
      primary_world: UInt64 0x0000000000006b27
      deleted_world: UInt64 0xffffffffffffffff
      specializations: Core.TypeMapEntry
        min_world: UInt64 0x0000000000000001
        max_world: UInt64 0xffffffffffffffff
  min_world: UInt64 0x0000000000000001
  max_world: UInt64 0xffffffffffffffff

julia> bar() = 2

julia> method_age(foo, Tuple{})
CodeInfo
  parent: MethodInstance
    def: Method
      primary_world: UInt64 0x0000000000006b26
      deleted_world: UInt64 0xffffffffffffffff
      specializations: Core.TypeMapEntry
        min_world: UInt64 0x0000000000000001
        max_world: UInt64 0xffffffffffffffff
  min_world: UInt64 0x0000000000000001
  max_world: UInt64 0xffffffffffffffff

julia> method_age(bar, Tuple{})
CodeInfo
  parent: MethodInstance
    def: Method
      primary_world: UInt64 0x0000000000006b28
      deleted_world: UInt64 0xffffffffffffffff
      specializations: Core.TypeMapEntry
        min_world: UInt64 0x0000000000000001
        max_world: UInt64 0xffffffffffffffff
  min_world: UInt64 0x0000000000000001
  max_world: UInt64 0xffffffffffffffff
```

PR #590, opened 2020-03-04
Move CUDA discovery back to the Pkg.build phase.

maleadt:
Going back and forth... Apparently we're not allowed to do significant work during init, so move CUDA discovery and artifact selection back to the Pkg build phase. This might complicate life for cluster users a little, but it still supports local CUDA discovery. For use on a login node where CUDA is not functional, that discovery process has been written in such a way that it does not require a GPU, but does require CUDA to be available at the final paths (i.e., the same ones as on compute nodes).


vchuravy:
:/ what is the particular issue triggered during init? I have been enjoying not having a build step. Additionally the depot can't be assumed to be writable (only the first entry in the depot list is required to)

maleadt:
I guess we could also move the discovery past init by having accessor functions for, e.g., nvdisasm, or the toolkit version, that lazily get a hold of the artifact or local installation when first called. That however will cause the download to happen even later, i.e. not during Pkg.build or when just loading the package, which might be a problem by itself on clusters: You'd have to make sure you call an actual API function on the login node. Isn't this about as complicated as having to run Pkg.build?

PR #162, opened 2018-02-13
Use new non-recursive codegen

maleadt:
Needs https://github.com/JuliaLang/julia/pull/25984, with additional changes from https://github.com/JuliaLang/julia/compare/jn/codegen-norecursion...tb/codegen-norecursion

PR #607, opened 2020-03-11
remove `cached` CodegenParams kwarg in >= v"1.5.0-DEV.393"

ianshmean:
My aim is to handle the removal of `cached` in https://github.com/JuliaLang/julia/commit/3b53f54a5e612cfc402f561d71cdbfb40da3529e#diff-ca03aefbdf6018c4da887b6ec35e0027

The specific build number might be wrong (I took yesterdays build number and subtracted 8).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -23,16 +23,16 @@ julia:1.3:
   variables:
     CI_THOROUGH: 'true'
 
-# julia:nightly:
-#   extends:
-#     - .julia:nightly
-#     - .test
-#   tags:
-#     - nvidia
-#     - sm_75
-#   variables:
-#     CI_THOROUGH: 'true'
-#   allow_failure: true
+julia:nightly:
+  extends:
+    - .julia:nightly
+    - .test
+  tags:
+    - nvidia
+    - sm_75
+  variables:
+    CI_THOROUGH: 'true'
+  allow_failure: true
 
 # julia:1.4-debug:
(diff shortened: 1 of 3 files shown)


vchuravy:
bors try

vchuravy:
the right way to test for that is to use the script in Julia `contrib/commit-name.sh` and pass it the commit hash that changed it.

ianshmean:
Build number updated. thanks

maleadt:
Well, this is https://github.com/JuliaGPU/CUDAnative.jl/pull/162 really and not expected to work by just removing `cached`. I'm surprised this passes tests...

ianshmean:
Ah ok. I hadn’t noticed that the nightly tests were commented out in the ci config, so it wasn’t being tested. 

I’ll close this. 

I am backed into a corner by this issue because I need the LLVM patch that is only present on julia master currently. I can try testing with the other PR given I don’t use float16

maleadt:
Feel fry to enable testing and try it out, it might work.

maleadt:
bors try

maleadt:
Oh yeah the hooks have changed and don't pass a worldcount anymore. So would need some more changed before we can actually test this.

maleadt:
I've made it so that https://github.com/JuliaGPU/CUDAnative.jl/pull/162 can be merged (retain compatibility with existing Julia, and conditionally disable tests that have been broken by the refactor).

PR #624, opened 2020-04-09
Make sure array intrinsics propagate inboundness.

maleadt:
Fixes https://github.com/JuliaGPU/CuArrays.jl/issues/677


maleadt:
There's some additional performance lost due to switching from an ad-hoc cartesian-like index tuple to actual Base.Cartesian, but I don't really feel like putting all that code back in...

PR #642, opened 2020-05-05
Avoid address space casts.

maleadt:
This PR changes how we work with and store address-space pointers: instead of casting them to and from generic pointers, directly work with their in-address space representation. That means we don't have to `addrspacecast` anymore, just `inttoptr ... addrspace(...)*` directly.

For WMMA, this requires 1.5's Base.AddrSpacePtr, because otherwise we select intrinsics for the generic case which leads to illegal memory accesses (cc @thomasfaingnaert).

Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/641.

### DIFF
--- a/src/device/cuda.jl
+++ b/src/device/cuda.jl
@@ -11,7 +11,9 @@ include("cuda/assertion.jl")
 include("cuda/memory_dynamic.jl")
 include("cuda/atomics.jl")
 include("cuda/misc.jl")
+if VERSION >= v"1.4.1"
 include("cuda/wmma.jl")
+end
 
 # functionality from libdevice
 #
--- a/src/device/cuda/memory_shared.jl
+++ b/src/device/cuda/memory_shared.jl
@@ -61,7 +61,6 @@ end
     end
 
     T_ptr = convert(LLVMType, DevicePtr{T,AS.Shared})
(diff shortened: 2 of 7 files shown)


thomasfaingnaert:
Will this completely eliminate all `addrspacecasts` in kernels? LLVM doesn't seem to handle them too well. In particular, I've noticed issues with the LSV and SeparateConstOffsetFromGEP passes.

For vectorisation in #629, I use the CUDA C++ approach of casting to `float4*` before loading/storing, which I'm not a big fan of. I suspect that, with this change, I can just call `unsafe_load` and `unsafe_store!` directly, so long as I use the correct alignment. The LSV pass will then handle the rest.

(Also, since `AddrSpacePtr` was backported to 1.4.1, the WMMA tests should still pass for non-debug 1.4.1 I believe, so 1.5 is too strict a requirement)

maleadt:
> Will this completely eliminate all `addrspacecasts` in kernels?

It should, yes.

> (Also, since `AddrSpacePtr` was backported to 1.4.1, the WMMA tests should still pass for non-debug 1.4.1 I believe, so 1.5 is too strict a requirement)

Ah yes, good call, it's only the return value that's problematic on 1.4.

